PR #6, opened 2014-05-16
Add support for #post_render_cache rendering of entities and cache tag support

davereid:
Requires the patch in https://drupal.org/node/2217877 to be applied to Drupal core.

### DIFF
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -78,15 +78,15 @@ class EntityEmbedFilter extends FilterBase {
         if (!empty($entity)) {
           $placeholder = $this->buildPlaceholder($entity, $view_mode, $langcode);
 
-          // Load the altered HTML into a new DOMDocument and retrieve the element.
+          // Load the placeholder HTML into a new DOMDocument and retrieve the element.
           $updated_node = Html::load($placeholder)->getElementsByTagName('body')
             ->item(0)
             ->childNodes
             ->item(0);
           // Import the updated node from the new DOMDocument into the original
-          // one, importing also the child nodes of the updated node.
+          // one, importing also the child nodes of the updated DOMNode.
           $updated_node = $dom->importNode($updated_node, TRUE);
-          // Finally, replace the original entity node with the new entity node!
+          // Replace the original entity DOMNode with the new entity DOMNode.
           $node->appendChild($updated_node);
         }
       }
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -148,17 +148,18 @@ class EntityEmbedFilter extends FilterBase {
         $entity_output = drupal_render($build);
 
         $depth--;
+
         $element['#markup'] = str_replace($placeholder, $entity_output, $element['#markup']);
         $element['#markup'] = str_replace($alt_placeholder, $entity_output, $element['#markup']);
-      }
-      else {
-        $element['#markup'] = str_replace($placeholder, '', $element['#markup']);
-        $element['#markup'] = str_replace($alt_placeholder, '', $element['#markup']);
+        return $element;
       }
     }
     catch (\Exception $e) {
       watchdog_exception('entity_embed', $e);
     }
+
+    $element['#markup'] = str_replace($placeholder, '', $element['#markup']);
+    $element['#markup'] = str_replace($alt_placeholder, '', $element['#markup']);
 
     return $element;
   }
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -26,28 +26,6 @@ use Drupal\filter\Plugin\FilterBase;
  */
 class EntityEmbedFilter extends FilterBase {
 
-  protected $postRender = array();
-  protected $cacheTags = array();
-
-  public function clearPostRender() {
-    $this->postRender = array();
-    $this->cacheTags = array();
-  }
-
-  public function buildPlaceholder($entity, $view_mode, $langcode) {
-    $callback = '\Drupal\entity_embed\Plugin\Filter\EntityEmbedFilter::postRender';
-    $context = array(
-      'entity_type' => $entity->getEntityTypeId(),
-      'entity_id' => $entity->id(),
-      'view_mode' => $view_mode,
-      'langcode' => $langcode,
-      'token' => drupal_render_cache_generate_token(),
-    );
-    $this->postRender[$callback][$context['token']] = $context;
-    $this->cacheTags[] = $entity->getCacheTag();
-    return drupal_render_cache_generate_placeholder($callback, $context, $context['token']);
-  }
-
   /**
    * {@inheritdoc}
    */
@@ -55,7 +33,13 @@ class EntityEmbedFilter extends FilterBase {
     if (stristr($text, 'data-entity-type') !== FALSE && stristr($text, 'data-view-mode') !== FALSE) {
       $dom = Html::load($text);
       $xpath = new \DOMXPath($dom);
-      $this->clearPostRender();
+
+      $build = array(
+        '#markup' => Html::serialize($dom),
+        '#post_render_cache' => array(),
+        '#cache' => array('tags' => array()),
+      );
+
       foreach ($xpath->query('//*[@data-entity-type and @data-view-mode]') as $node) {
         $entity_type = $node->getAttribute('data-entity-type');
         $entity = NULL;
@@ -97,10 +81,6 @@ class EntityEmbedFilter extends FilterBase {
           $node->appendChild($updated_node);
         }
       }
-
-      $build = array(
-        '#markup' => Html::serialize($dom),
-      );
       if (!empty($this->postRender)) {
         $build['#post_render_cache'] = $this->postRender;
       }
@@ -128,6 +108,20 @@ class EntityEmbedFilter extends FilterBase {
     else {
       return $this->t('You can embed entities.');
     }
+  }
+
+  public function buildPlaceholder($entity, $view_mode, $langcode, array &$build) {
+    $callback = '\Drupal\entity_embed\Plugin\Filter\EntityEmbedFilter::postRender';
+    $context = array(
+      'entity_type' => $entity->getEntityTypeId(),
+      'entity_id' => $entity->id(),
+      'view_mode' => $view_mode,
+      'langcode' => $langcode,
+      'token' => drupal_render_cache_generate_token(),
+    );
+    $build['#post_render_cache'][$callback][$context['token']] = $context;
+    $build['#cache']['tags'] = NestedArray::mergeDeepArray($build['#cache']['tags'], $entity->getCacheTag());
+    return drupal_render_cache_generate_placeholder($callback, $context, $context['token']);
   }
 
   public static function postRender(array $element, array $context) {
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -8,6 +8,7 @@
 namespace Drupal\entity_embed\Plugin\Filter;
 
 use Drupal\Component\Utility\Html;
+use Drupal\Component\Utility\NestedArray;
 use Drupal\Component\Utility\String;
 use Drupal\Component\Utility\Unicode;
 use Drupal\Component\Utility\Xss;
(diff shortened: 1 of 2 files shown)


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -20,19 +21,41 @@
...
+    $callback = '\Drupal\entity_embed\Plugin\Filter\EntityEmbedFilter::postRender';
+    $context = array(
+      'entity_type' => $entity->getEntityTypeId(),
+      'entity_id' => $entity->id(),
+      'view_mode' => $view_mode,
+      'langcode' => $langcode,
+      'token' => drupal_render_cache_generate_token(),
+    );
+    $this->postRender[$callback][$context['token']] = $context;
+    $this->cacheTags[] = $entity->getCacheTag();
```
wimleers: This is wrong because `::buildPlaceHolder()` will _not_ be called in case of a cache hit! You must load the entity in the `#post_render_cache` callback and then do this. You need to load the entity there already anyway, so it's trivial to remove this here and add it there.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -20,19 +21,41 @@
...
+
   /**
    * {@inheritdoc}
    */
-  public function process($text, $langcode, $cache, $cache_id) {
+  public function process($text, $langcode) {
     if (stristr($text, 'data-entity-type') !== FALSE && stristr($text, 'data-view-mode') !== FALSE) {
       $dom = Html::load($text);
       $xpath = new \DOMXPath($dom);
+      $this->clearPostRender();
```
wimleers: For the reasons cited above, this can and should also be removed, along with the implementation of `::clearPostRender()`.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -52,40 +75,34 @@ public function process($text, $langcode, $cache, $cache_id) {
...
-          $updated_node = Html::load($output)->getElementsByTagName('body')
+          $updated_node = Html::load($placeholder)->getElementsByTagName('body')
             ->item(0)
             ->childNodes
             ->item(0);
           // Import the updated node from the new DOMDocument into the original
           // one, importing also the child nodes of the updated node.
           $updated_node = $dom->importNode($updated_node, TRUE);
-          // Finally, replace the original image node with the new image node!
+          // Finally, replace the original entity node with the new entity node!
```
wimleers: "Entity node"… :P 
You might want to reword that :)

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -52,40 +75,34 @@ public function process($text, $langcode, $cache, $cache_id) {
...
           $node->appendChild($updated_node);
-
-          return Html::serialize($dom);
         }
       }
+
+      $text = Html::serialize($dom);
+
+      if (!empty($this->postRender) || !empty($this->cacheTags)) {
+        $return = array(
```
wimleers: We typically call this `$build`.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -52,40 +75,34 @@ public function process($text, $langcode, $cache, $cache_id) {
...
+          // Finally, replace the original entity node with the new entity node!
           $node->appendChild($updated_node);
-
-          return Html::serialize($dom);
         }
       }
+
+      $text = Html::serialize($dom);
+
+      if (!empty($this->postRender) || !empty($this->cacheTags)) {
```
wimleers: _Every_ entity has (_must have_) cache tags, so the second condition should always evaluate to true, hence the entire expression should always evaluate to true.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -107,4 +124,46 @@ public function tips($long = FALSE) {
...
+    // to be sure to check for both formats of the render cache placeholder:
+    // Original placeholder:
+    // <drupal:render-cache-placeholder .. />
+    // After FilterHtmlCorrector::process():
+    // <render-cache-placeholder ... ></render-cache-placeholder>
+    $alt_placeholder = Html::normalize($placeholder);
+
+    $entity = entity_load($context['entity_type'], $context['entity_id']);
+    if ($entity && $entity->access()) {
+      // Protect ourselves from recursive rendering.
```
wimleers: This should not be necessary.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -107,4 +124,46 @@ public function tips($long = FALSE) {
...
+      $depth++;
+      if ($depth > 10) {
+        throw new RecursiveRenderingException(format_string('Recursive rendering detected when rendering entity @entity_type(@entity_id). Aborting rendering.', array('@entity_type' => $item->entity->getEntityTypeId(), '@entity_id' => $item->target_id)));
+      }
+
+      // Build the rendered entity.
+      $build = entity_view($entity, $context['view_mode'], $context['langcode']);
+
+      // Hide entity links by default.
+      // @todo Make this configurable via data attribute?
```
wimleers: Yes! :)

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: I'm confused. We're passing in the entity info and doing the loading and access checking in the post-render cache callback itself?

davereid: It's already understood this code only executes in the case of a cache miss. It's only going to add the placeholder and let post_render_cache magic do the rest.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: Not understood.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: Hey, blame FilterCaption, it says "image node" in it's inline docs. :)

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: The post render cache callback isn't necessary?

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: Consider an embedded entity is deleted but we still attempt to render the embed code - these would both be empty in that case.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: No, the protection against recursive rendering shouldn't be necessary.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: Aha! :) Interesting. Sounds good.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: But "image" is not an entity type! :D

davereid: The context is the same as FilterCaption: we're working with an DOM node, not a Drupal node. The node represents the embedded entity HTML.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: Node 1 embeds Node 2 in its body field. Node 2 embeds Node 1 in its body field. _boom_? I guess I'll try and see what happens now that we're using the post_render_cache.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: Ideally we should be able to control links display via 'Manage display' - I hate hard-coding it here or making it an option.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: You're right, of course, I was just pointing out that this could be confusing :)

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: Ah wow, hadn't considered that.

In that case, just using `static $depth++` is wrong, because `::postRender()` will be called once per entity. If you are rendering a page that with filtered text that has 11 entities to be rendered, then the 11th won't be rendered, even though no recursion is involved.

Not sure yet how you'd detect that.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: Ah, I thought this was because, for example, even though a node in the "teaser" view mode needs to have links, you don't want those links to be rendered when embedded in an entity.

You're right that links should be configurable in entity view displays though.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: I just got thrown off by how you store data outside of the current scope. I thought you were storing the cache tags here to be set by the `#post_render_cache` callback. But that's not the case.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
slashrsm: There is a $depth-- a bit further down that should prevent that from happening.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: Yeah, I'd love to have an 'Embed' view mode enabled by default (which has links disabled) for all entity types - which D8 core makes really hard to do now that they're managed in config.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: Same as above:

> I just got thrown off by how you store data outside of the current scope. I thought you were storing the cache tags here to be set by the `#post_render_cache` callback. But that's not the case.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: Step 1 would be to make `links` a display component like any other. There probably is already an issue for this.

Then, you can either advocate to get this in core (I'll help), or you could generate an `embed` view display automatically upon installation; you could read the default view display for every entity type, remove the `links` component, if any, and then save that as your `embed` view display :)

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
wimleers: I think you're right. Seems like _all_ my remarks are wrong :P Just trying to help! :)

davereid: @wimleers We're still very thankful though!

slashrsm: @wimleers++

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
slashrsm: +1 for fixing this in core.

davereid: Filed https://drupal.org/node/2271349

davereid:
Would love to get an OK from both @slashrsm and @cs-shadow before merging this PR.

slashrsm:
I am fine with it.

cs-shadow:
Looks good to me too.

PR #7, opened 2014-05-16
Entity UUID preferred over ID and comment clarifications.

cs-shadow:


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -38,12 +38,12 @@ public function process($text, $langcode, $cache, $cache_id) {
...
         $view_mode = $node->getAttribute('data-view-mode');
 
-        // Load the entity either by UUID or ID.
+        // Load the entity either by UUID (preferred) or ID.
         if ($node->hasAttribute('data-entity-uuid')) {
           $uuid = $node->getAttribute('data-entity-uuid');
           $entity = entity_load_by_uuid($entity_type, $uuid);
         }
-        if ($node->hasAttribute('data-entity-id')) {
+        else if ($node->hasAttribute('data-entity-id')) {
```
davereid: &lt;nitpick&gt;According to coding standards, should be `elseif` &lt;/nitpick&gt;

cs-shadow: Fixed in latest commit.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -80,7 +81,12 @@ public function process($text, $langcode, $cache, $cache_id) {
           // Import the updated node from the new DOMDocument into the original
           // one, importing also the child nodes of the updated node.
           $updated_node = $dom->importNode($updated_node, TRUE);
-          // Finally, replace the original image node with the new image node!
+
+          // Remove all children of the node from the existing DOMDocument.
```
davereid: I'm just curious about this change. Why?

cs-shadow: As discussed in this issue: https://drupal.org/node/2269135, we remove any existing children of the DOM node. That's treated as alt text or a placeholder and hence removed before embedding.

PR #8, opened 2014-05-16
Embeds replace existing content instead of appending.

cs-shadow:
Based on the issue: https://drupal.org/node/2269135

Note: This PR works on top of the PR #7.

### DIFF
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -80,6 +80,11 @@ class EntityEmbedFilter extends FilterBase {
           // Import the updated node from the new DOMDocument into the original
           // one, importing also the child nodes of the updated node.
           $updated_node = $dom->importNode($updated_node, TRUE);
+
+          // Remove all children of the node from the existing DOMDocument.
+          while ($node->hasChildNodes() == TRUE) {
+            $node->removeChild($node->firstChild);
+          }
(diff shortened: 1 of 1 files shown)


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -80,7 +81,12 @@ public function process($text, $langcode, $cache, $cache_id) {
           // Import the updated node from the new DOMDocument into the original
           // one, importing also the child nodes of the updated node.
           $updated_node = $dom->importNode($updated_node, TRUE);
-          // Finally, replace the original image node with the new image node!
+
+          // Remove all children of the node from the existing DOMDocument.
+          while ($node->hasChildNodes() == TRUE) {
+            $node->removeChild($node->firstChild);
```
davereid, line 87: If we had content like this would this still work?

```
<div data-entity-type="node" data-entity-id="1" data-view-mode="embed">
  <p>First paragraph<p>
  <p>Second paragraph</p>
</div>
```

Or would it only replace the 'First paragraph' element?

cs-shadow, line 87: It'll work because there's a while loop running that checks if the node has any child nodes left. As far as I've tested it, it works. I'll soon provide test coverage for the same.

davereid, line 87: Ah thanks, yes, I completely missed that it was a while loop!

PR #9, opened 2014-05-17
Remove all children of the node from the existing DOMDocument.

cs-shadow:


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -82,6 +82,10 @@ public function process($text, $langcode, $cache, $cache_id) {
           // one, importing also the child nodes of the updated node.
           $updated_node = $dom->importNode($updated_node, TRUE);
 
+          // Remove all children of the node from the existing DOMDocument.
+          while ($node->hasChildNodes() == TRUE) {
```
davereid: Does hasChildNodes() return anything other than a boolean? If not, we can just skip the == TRUE part.

cs-shadow: It indeed returns only boolean values. I'll remove the ==TRUE part in next commit, once I'm done with the tests for this.

PR #10, opened 2014-05-20
Check whether the specified entity_type exists before trying to embed entity.

cs-shadow:
Details in drupal.org issue: https://drupal.org/node/2271185.

Should be merged after PR #9 passes through.

### DIFF
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -83,7 +83,7 @@ class EntityEmbedFilter extends FilterBase {
           $updated_node = $dom->importNode($updated_node, TRUE);
 
           // Remove all children of the node from the existing DOMDocument.
-          while ($node->hasChildNodes() == TRUE) {
+          while ($node->hasChildNodes()) {
             $node->removeChild($node->firstChild);
           }
           // Finally, append the entity to the DOM node.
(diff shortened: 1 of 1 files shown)


davereid:
I think it would probably be better to wrap all the entity loading and rendering using a `try { ... } catch (Exception $e) { ... }` statement since if entity loading fails (due to not valid entity type), or entity rendering fails, it will catch all possible errors.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -82,6 +89,10 @@ public function process($text, $langcode, $cache, $cache_id) {
           // one, importing also the child nodes of the updated node.
           $updated_node = $dom->importNode($updated_node, TRUE);
 
+          // Remove all children of the node from the existing DOMDocument.
```
slashrsm, line 92: This belongs to another PR. You should separate things.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -38,6 +38,13 @@ public function process($text, $langcode, $cache, $cache_id) {
         $entity = NULL;
         $view_mode = $node->getAttribute('data-view-mode');
 
+        // Continue only if specified entity type exists.
+        $entity_info = \Drupal::entityManager()->getDefinition($entity_type);
+        if($entity_info == NULL) {
+          watchdog($entity_type, "Specified entity type '%entity_type' does not exists.", array('%entity_type' => $entity_type));
```
slashrsm, line 44: Let's use "entity_embed" instead of $entity_type. This will clearly show that this error happened here.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid, line 92: Yeah, it looks like you're creating new PRs from existing PR branches, which makes things difficult to review. It would help to always create new branches/PRs from 8.x-1.x.

cs-shadow, line 92: My bad. Remnant of previous PR.

PR #12, opened 2014-05-20
All entity loading and rendering wrapped into try-catch block.

cs-shadow:
Details in drupal.org issue: https://drupal.org/node/2271185.

### DIFF
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -94,7 +94,7 @@ class EntityEmbedFilter extends FilterBase {
           }
         }
         catch(\Exception $e) {
-          watchdog_exception($entity_type, $e);
+          watchdog_exception('entity_embed', $e);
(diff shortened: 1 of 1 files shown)


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -38,58 +38,63 @@ public function process($text, $langcode, $cache, $cache_id) {
...
+            }
+            // Finally, append the entity to the DOM node.
+            $node->appendChild($updated_node);
 
-          return Html::serialize($dom);
+            return Html::serialize($dom);
+          }
+        }
+        catch(\Exception $e) {
+          watchdog_exception($entity_type, $e);
```
davereid: Like the previous PR, let's use 'entity_embed' instead of `$entity_type` here for the first parameter.

cs-shadow: Sorry, forgot to change that. Fixed in latest commit.

PR #13, opened 2014-05-20
Control display of entity links using data-links-show attribute. Hide by default.

davereid:


cs-shadow:
Okay, then its good to go.

slashrsm:
Looks good. Conflicts on merge.

cs-shadow:
Conflicts resolved in PR #16

davereid:
I can reroll this rather than needing a separate PR.

cs-shadow:
Sure, go ahead.

davereid:
Postponing on https://github.com/drupal-media/entity_embed/pull/15 which tweaks data attributes and the rendering process.

davereid:
I think this should be closed in favor of https://drupal.org/node/2273393 which would allow the 'links' to be configurable via the entity reference formatter.

davereid:
https://drupal.org/node/2274975 is further proof that the current method just doesn't work well and I'd rather rely on the entity reference formatter to handle that logic for us.

davereid:
I'm going to close this for now. In light of us being able to use field formatters now with the display plugins, I don't think it makes sense to add a data attribute for this. We can just support this via the entity reference field formatter options, or whatever it ends up being.

PR #14, opened 2014-05-21
Store entity manager controller as a class member and use it to load/render entities

cs-shadow:
Details in this issue: https://drupal.org/node/2271203

### DIFF
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -161,7 +161,8 @@ class EntityEmbedFilter extends FilterBase implements ContainerFactoryPluginInte
     $alt_placeholder = Html::normalize($placeholder);
 
     try {
-      $entity = entity_load($context['entity_type'], $context['entity_id']);
+      $controller = $this->entity_manager->getStorage($context['entity_type']);
+      $entity = $controller->load($context['entity_id']);
       if ($entity && $entity->access()) {
         // Protect ourselves from recursive rendering.
         static $depth = 0;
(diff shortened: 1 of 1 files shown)


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -42,11 +69,18 @@ public function process($text, $langcode, $cache, $cache_id) {
           // Load the entity either by UUID (preferred) or ID.
           if ($node->hasAttribute('data-entity-uuid')) {
             $uuid = $node->getAttribute('data-entity-uuid');
-            $entity = entity_load_by_uuid($entity_type, $uuid);
+
+            $entity_type_definition = \Drupal::entityManager()->getDefinition($entity_type);
```
davereid: Shouldn't this use $this->entity_manager?

cs-shadow: It should be. My bad. Fixed in latest commit.

slashrsm:
Looks good but apparently there are merge conflicts. Please re-roll and it should be ready to go in.

cs-shadow:
Re-rolled against latest HEAD.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -137,7 +172,9 @@ public static function postRender(array $element, array $context) {
         }
 
         // Build the rendered entity.
-        $build = entity_view($entity, $context['view_mode'], $context['langcode']);
+        $entity_type_id = $entity->getEntityTypeId();
+        $render_controller = $this->entity_manager->getViewBuilder($entity_type_id);
```
davereid, line 176: This was not actually tested and does not work. This is a public static function and does not have a $this-> available.

davereid, line 176: I had to revert the changes in postRender() for now otherwise the module wasn't even working.

cs-shadow, line 176: Sorry for the trouble. I forgot to clear the cache while testing and it worked. After clearing all caches it gives on error.

PR #15, opened 2014-05-22
Added alter hooks for embed placeholder context and for the rendered embedded entity

davereid:
I realized other modules will probably want to add extra data attributes via the embed form, and have them affect entity display. Now that we're using post_render_cache we lose a bit of that context when doing the actual rendering. My proposal is for the following:
- Add all the attributes of the embed element into the post_render_cache token. So data-entity-type becomes `$context['entity-type']`.
- Add an alter hook for the $context array before it is converted into a placeholder.
- Add an alter hook or some kind of property to the entity before calling entity_view() to indicate that it is an embedded entity.

### DIFF
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -103,15 +103,15 @@ class EntityEmbedFilter extends FilterBase implements ContainerFactoryPluginInte
           if ($node->hasAttribute('data-entity-uuid')) {
             $uuid = $node->getAttribute('data-entity-uuid');
 
-            $entity_type_definition = $this->entity_manager->getDefinition($entity_type);
+            $entity_type_definition = $this->entityManager->getDefinition($entity_type);
             $uuid_key = $entity_type_definition->getKey('uuid');
-            $controller = $this->entity_manager->getStorage($entity_type);
+            $controller = $this->entityManager->getStorage($entity_type);
             $entity = reset($controller->loadByProperties(array($uuid_key => $uuid)));
           }
           elseif ($node->hasAttribute('data-entity-id')) {
             $id = $node->getAttribute('data-entity-id');
 
-            $controller = $this->entity_manager->getStorage($entity_type);
+            $controller = $this->entityManager->getStorage($entity_type);
             $entity = $controller->load($id);
 
             // Add the entity UUID attribute to the parent node.
--- a/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
+++ b/lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
@@ -12,6 +12,7 @@ use Drupal\Component\Utility\NestedArray;
 use Drupal\Component\Utility\String;
 use Drupal\Component\Utility\Unicode;
 use Drupal\Component\Utility\Xss;
+use Drupal\Core\Language\Language;
 use Drupal\filter\Plugin\FilterBase;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -93,15 +100,19 @@ public function tips($long = FALSE) {
...
+    $context += array(
+      'entity-type' => $entity->getEntityTypeId(),
+      'entity-id' => $entity->id(),
+      'view-mode' => 'default',
+      'langcode' => Language::LANGCODE_DEFAULT,
     );
+    $context['token'] = drupal_render_cache_generate_token();
+
+    // Allow modules to alter the context.
+    \Drupal::moduleHandler()->alter('entity_embed_context', $context, $entity);
```
slashrsm: Let's inject module handler via DIC. Also, would it make sense to emmit an event here? We could then implement hooks in event subscriber if needed.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -145,6 +157,9 @@ public static function postRender(array $element, array $context) {
           $build['links']['#access'] = FALSE;
         }
 
+        // Allow modules to alter the rendered embedded entity.
+        \Drupal::moduleHandler()->alter('entity_embed', $build, $entity, $context);
```
slashrsm: As above.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: Yep, definitely will convert now the the module handler DIC since we're using it now.

I don't think it makes a lot of sense to emit an event instead of hooks here. All of D8's entity rendering alterations still rely on hooks - [hook_entity_view_mode_alter](https://api.drupal.org/api/drupal/core%21modules%21system%21entity.api.php/function/hook_entity_view_mode_alter/8), [hook_entity_view](https://api.drupal.org/api/drupal/core%21modules%21system%21entity.api.php/function/hook_entity_view/8), and [hook_entity_view_alter](https://api.drupal.org/api/drupal/core%21modules%21system%21entity.api.php/function/hook_entity_view_alter/8). And unlike the post_render_cache patch, which was actively impeding performance and progress, I think we can safely wait until https://drupal.org/node/1972304 has landed to worry about Events.

slashrsm: I know events are quite controversial topic, but I honestly think they're much more convenient to use in OO code than hooks. That's why I see them being used more and more when people really start to work on D8. 

If emmiting an event here and implementing subscriber that fires hooks doesn't work, then I'd at least like to see both being fired here. However, I still think first solution is much cleaner.

davereid: I feel strongly that this is not worth diverging from the standards set by core. There is a clear established pattern in core that nothing with entity rendering alteration uses Events.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
davereid: This is a public static method. We do not have access to any injected dependencies here.

lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
slashrsm: > There is a clear established pattern in core that nothing with entity rendering alteration uses Events.

Is this documented anywhere?

I'd say let's merge this and decide on approach with events in a separate ticket.

PR #19, opened 2014-05-23
Added missing type hinting for  in setDomNodeContent.

cs-shadow:
Details here: https://drupal.org/node/2272747


lib/Drupal/entity_embed/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -197,7 +197,7 @@ public static function postRender(array $element, array $context) {
   /**
    * Replace the contents of a DOMNode.
    *
-   * @param $node
+   * @param \DOMNode|\DOMElement $node
```
davereid: I think we're ok to do just `\DOMNode` here since DOMElement is a subclass of DOMNode.

davereid:
We should add the actual `\DOMNode $node` type hint to the actual function parameter as well.

cs-shadow:
Will do that.

PR #20, opened 2014-05-23
Access operation is required

webflo:
The "operation" is a required by the AccessibleInterface::access.


davereid:
Oh strange. Node::access() provides the default of 'view' but ContentEntityBase::access() does not have a default. That almost sounds like we should file a core bug about that.

webflo:
Yes, this is valid in terms of OOP inheritance but `Node::access` is the only entity type in core with a default parameter.

PR #21, opened 2014-05-29
Abstract entity rendering with EntityEmbedDisplay plugin type.

davereid:
I think this is how I intend to solve https://drupal.org/node/2273393. With an abstract EntityEmbedDisplay, I could add sub-classes for Entity Reference formatters, or other field formatters. This would allow users to create arbitrary ways to display embedded entities as well outside of just entity_view(), which will be very useful.

Tasks:
- [x] Convert to plugins
- [x] Figure out if derivative plugins can work for field formatter plugins - yes they work!
- [x] Integrate with PluginFormInterface
- [x] Ensure all new methods and classes are documented.
- [x] Updated docs about new data attributes.

### DIFF
--- a/entity_embed.api.php
+++ b/entity_embed.api.php
@@ -11,7 +11,7 @@
  */
 
 // @todo Document
-function hook_entity_embed_display_info_alter(&$info) {
+function hook_entity_embed_display_plugins_alter(&$info) {
 
 }
 
--- a/src/Annotation/EntityEmbedDisplay.php
+++ b/src/Annotation/EntityEmbedDisplay.php
@@ -37,13 +37,11 @@ class EntityEmbedDisplay extends Plugin {
   /**
    * The entity types the display can apply to.
    *
-   * To make the display plugin valid for all entity types, include the string
-   * 'entity' in the types.
+   * To make the display plugin valid for all entity types, set this value to
+   * FALSE.
    *
-   * @todo Replace with \Drupal\Core\Plugin\Context\Context?
-   *
-   * @var array
+   * @var bool|array
    */
-  public $types = array();
+  public $entity_types = FALSE;
 
 }
--- a/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
@@ -7,8 +7,6 @@
 
 namespace Drupal\entity_embed\EntityEmbedDisplay;
 
-use Drupal\entity_embed\EntityEmbedDisplayBase;
-
 /**
  * Default embed display, which renders the entity using entity_view().
  *
--- a/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
@@ -2,10 +2,10 @@
 
 /**
  * @file
- * Contains \Drupal\entity_embed\Plugin\EntityEmbedDisplay\EntityEmbedDefaultDisplay.
+ * Contains \Drupal\entity_embed\EntityEmbedDisplay\EntityEmbedDefaultDisplay.
  */
 
-namespace Drupal\entity_embed\Plugin\EntityEmbedDisplay;
+namespace Drupal\entity_embed\EntityEmbedDisplay;
 
 use Drupal\entity_embed\EntityEmbedDisplayBase;
 
--- a/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
@@ -49,13 +49,12 @@ class EntityEmbedDefaultDisplay extends EntityEmbedDisplayBase {
    * {@inheritdoc}
    */
   public function build() {
-    // Clone the entity since we're going to set some additional properties we
-    // don't want kept around afterwards.
-    $entity = clone $this->entity;
-    $entity->entity_embed_attributes = $this->getAttributes();
+    $entity = $this->getContextValue('entity');
+    $view_mode = $this->getConfigurationValue('view_mode');
+    $langcode = $this->getAttributeValue('langcode');
 
     // Build the rendered entity.
-    $build = entity_view($this->entity, $this->getConfigurationValue('view_mode'), $this->getAttributeValue('langcode'));
+    $build = entity_view($entity, $view_mode, $langcode);
 
     // Hide entity links by default.
     if (isset($build['links'])) {
--- a/src/EntityEmbedDisplay/EntityEmbedDisplayBase.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDisplayBase.php
@@ -58,7 +58,7 @@ abstract class EntityEmbedDisplayBase extends PluginBase implements EntityEmbedD
   }
 
   /**
-   * Validate that this display plugin applies to the current entity type.
+   * Validates that this display plugin applies to the current entity type.
    *
    * This checks the plugin annotation's 'entity_types' value, which should be
    * an array of entity types that this plugin can process, or FALSE if the
--- a/src/EntityEmbedDisplay/EntityEmbedDisplayBase.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDisplayBase.php
@@ -85,28 +85,28 @@ abstract class EntityEmbedDisplayBase extends PluginBase implements EntityEmbedD
   abstract public function build();
 
   /**
-   * Implements \Drupal\Component\Plugin\ConfigurablePluginInterface::calculateDependencies().
+   * {@inheritdoc}
    */
   public function calculateDependencies() {
     return array();
   }
 
   /**
-   * Implements \Drupal\Component\Plugin\ConfigurablePluginInterface::defaultConfiguration().
+   * {@inheritdoc}
    */
   public function defaultConfiguration() {
     return array();
   }
 
   /**
-   * Implements \Drupal\Component\Plugin\ConfigurablePluginInterface::getConfiguration().
+   * {@inheritdoc}
    */
   public function getConfiguration() {
     return $this->configuration;
   }
 
   /**
-   * Implements \Drupal\Component\Plugin\ConfigurablePluginInterface::setConfiguration().
+   * {@inheritdoc}
    */
   public function setConfiguration(array $configuration) {
     $this->configuration = NestedArray::mergeDeep(
@@ -117,21 +117,21 @@ abstract class EntityEmbedDisplayBase extends PluginBase implements EntityEmbedD
   }
 
   /**
-   * Implements \Drupal\Component\Plugin\ConfigurablePluginInterface::buildConfiguration().
+   * {@inheritdoc}
    */
   public function buildConfigurationForm(array $form, array &$form_state) {
     return array();
   }
 
   /**
-   * Implements \Drupal\Component\Plugin\ConfigurablePluginInterface::validateConfigurationForm().
+   * {@inheritdoc}
    */
   public function validateConfigurationForm(array &$form, array &$form_state) {
     // Do nothing.
   }
 
   /**
-   * Implements \Drupal\Component\Plugin\ConfigurablePluginInterface::submitConfigurationForm().
+   * {@inheritdoc}
    */
   public function submitConfigurationForm(array &$form, array &$form_state) {
     if (!form_get_errors($form_state)) {
--- a/src/EntityEmbedDisplay/EntityEmbedDisplayBase.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDisplayBase.php
@@ -41,6 +41,18 @@ abstract class EntityEmbedDisplayBase extends PluginBase implements EntityEmbedD
     $this->setConfiguration($configuration);
   }
 
+  /**
+   * Gets a configuration value.
+   *
+   * @param string $name
+   *   The name of the plugin configuration value.
+   * @param mixed $default
+   *   The default value to return if the configuration value does not exist.
+   *
+   * @return mixed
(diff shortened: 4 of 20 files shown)


davereid:
Here's what this pull-request makes possible:

5387f8c6c38aa54752bb6c3f

5387f8dcc38aa547518e072f

5388964ec38aa5475660b4de

slashrsm:
~~Could we design this in a way that we wouldn't need to implement glue class for every possible formatter?~~

EDIT: Ah... forget it.

src/Annotation/EntityEmbedDisplay.php
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @var \Drupal\Core\Annotation\Translation
+   */
+  public $label = '';
+
+  /**
+   * The entity types the display can apply to.
+   *
+   * To make the display plugin valid for all entity types, include the string
+   * 'entity' in the types.
```
slashrsm: Could we achieve same thing in a less hacky way?

src/Plugin/Derivative/FieldFormatter.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Drupal\entity_embed\Plugin\Derivative;
+
+use Drupal\Component\Plugin\Derivative\DerivativeBase;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Plugin\Discovery\ContainerDerivativeInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Field\FormatterPluginManager;
+
+/**
+ * Provides block plugin definitions for custom menus.
```
slashrsm: Leftover.

src/Annotation/EntityEmbedDisplay.php
davereid: I'm open to suggestions. This seemed to be the easiest since it's easy to define which entity types should be allowed to be rendered with a display plugin just using the annotation.

src/FieldFormatterEntityEmbedDisplayBase.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\entity_embed\FieldFormatterEntityEmbedDisplayBase.
+ */
+
+namespace Drupal\entity_embed;
+
+use Drupal\entity_embed\EntityEmbedDisplayBase;
```
jibran: We can remove this.

src/EntityEmbedDisplayBase.php
```diff
@@ -0,0 +1,120 @@
...
+ * @file
+ * Contains \Drupal\entity_embed\EntityEmbedDisplayBase.
+ */
+
+namespace Drupal\entity_embed;
+
+use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Plugin\PluginBase;
+use Drupal\entity_embed\EntityEmbedDisplayInterface;
```
jibran: This can be removed.

davereid: I don't understand how. It's used in the 'implements' statement of the class definition?

src/FieldFormatterEntityEmbedDisplayBase.php
davereid: I don't understand how. It's used in the 'extends' statement of the class definition?

jibran: I mean they have the same namespace. According to http://www.php.net/manual/en/language.namespaces.rules.php it always try to look `EntityEmbedDisplayBase` in `Drupal\entity_embed` namespace first if we remove `use Drupal\entity_embed\EntityEmbedDisplayBase;`.
This is extreme nitpick. :)

src/EntityEmbedDisplayBase.php
jibran: same as above.

src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -120,10 +120,26 @@ public function process($text, $langcode) {
...
+              // Check for JSON-encoded attributes.
+              $data = json_decode($context[$key], TRUE, 10);
+              if ($data !== NULL && json_last_error() === JSON_ERROR_NONE) {
+                $context[$key] = $data;
+              }
+            }
+
+            // Support the deprecated view-mode data attribute.
+            if (isset($context['view-mode']) && !isset($context['entity-embed-display']) && !isset($context['entity-embed-settings'])) {
+              $context['entity-embed-settings']['view-mode'] = $context['view-mode'];
```
slashrsm: This doesn't seem to work (tested it). Should we set entity-embed-display here?

slashrsm:
It would be great if we'd add a README.txt and document data attributes that are added with this PR. It seems that things get a bit more complicated with this approach and some docs would probably help other people to understand it.

davereid:
@slashrsm Yep, updating docs (README.txt) is an open task to be done in the first post above.

slashrsm:
++

src/Plugin/Filter/EntityEmbedFilter.php
davereid: The buildPlaceholder() method should provide a default of `$context['entity-embed-display'] = 'default'`. Is that not actually happening?

slashrsm: Yes... you are right. Ignore it.

src/Annotation/EntityEmbedDisplay.php
slashrsm: At least a wildcard ('*') if nothing else. I'm not familiar that much with Conditions in D8, but I'd say that would be the preferred way.

davereid: I just talked with @timplunkett about the generic way to handle 'restrictions' like this and came to the conclusion that having plugin contexts is the way to go here. It means that plugins that restrict by entity type would be limited to only one entity type. Based on what Tim has also been working on, it seems the context system needs a bit of work. I'll look into converting this to see if it's feasible currently.

davereid: There are way too many obsticles to using the context system for this. I've actually found that core uses this same pattern in the annotation:

Constraint plugins restricted by typed data type:

```
/**
 * Validation constraint for the entity changed timestamp.
 *
 * @Plugin(
 *   id = "EntityChanged",
 *   label = @Translation("Entity changed", context = "Validation"),
 *   type = {"entity"}
 * )
 */
```

Field Formatter and Widget plugins use the same thing.

I think rather than keying based on 'entity' that if the 'type' value isn't in the annotation, then it is assumed to be allowed for all entity types.

slashrsm: Yah, I agree. That seems nicer.

slashrsm: Should we create a follow-up about introducing context at some later stage?

davereid:
Ok I believe this is fully ready to review and merge. I've added docs to all methods, and included docs in the README about the plugin system usage with the embed code.

entity_embed.api.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+/**
+ * @addtogroup hooks
+ * @{
+ */
+
+// @todo Document
+function hook_entity_embed_display_plugins_alter(&$info) {
+
```
jibran, line 15: Some doc and code example missing.

entity_embed.api.php
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @param array &$context
+ *   The context array.
+ * @param callable &$callback
+ *   The post-render callback to be used.
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   The entity being rendered.
+ */
+function hook_entity_embed_context_alter(array &$context, &$callback, \Drupal\Core\Entity\EntityInterface $entity) {
+
```
jibran, line 29: Perhaps some code example.

src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $entity = $this->getContextValue('entity');
+    $view_mode = $this->getConfigurationValue('view_mode');
+    $langcode = $this->getAttributeValue('langcode');
+
+    // Build the rendered entity.
+    $build = entity_view($entity, $view_mode, $langcode);
```
jibran, line 57: We can inject \Drupal::entityManager() and then use $render_controller->view() instead see https://api.drupal.org/api/drupal/core!includes!entity.inc/function/entity_view/8

src/EntityEmbedDisplay/EntityEmbedDisplayBase.php
```diff
@@ -0,0 +1,228 @@
...
+    if (!$this->isValidEntityType()) {
+      return FALSE;
+    }
+
+    // Check that the entity itself can be viewed by the user.
+    return $this->getContextValue('entity')->access('view', $account);
+  }
+
+  /**
+   * Validate that this display plugin applies to the current entity type.
```
jibran: Validates

src/EntityEmbedDisplay/EntityEmbedDisplayInterface.php
```diff
@@ -0,0 +1,44 @@
...
+use Drupal\Component\Plugin\ConfigurablePluginInterface;
+use Drupal\Core\Plugin\PluginFormInterface;
+use Drupal\Core\Session\AccountInterface;
+
+/**
+ * Defines the required interface for all entity embed display plugins.
+ *
+ * @ingroup entity_embed_api
+ */
+interface EntityEmbedDisplayInterface extends ConfigurablePluginInterface, PluginFormInterface, PluginInspectionInterface {
```
jibran, line 20: Why all the public methods in `EntityEmbedDisplayBase` are not part of this interface other then public methods defined in `PluginBase`?

jibran:
Very nice work @davereid this is looking really awesome.
There are a lot of implements and overrides doc blocks. I think we should replace it with `{@inheritdoc}`. I am just trying to help here so please discard whatever feedback is irreverent. :)

src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -86,11 +86,11 @@ public static function create(ContainerInterface $container, array $configuratio
...
-    if (strpos($text, 'data-entity-type') !== FALSE && strpos($text, 'data-view-mode') !== FALSE) {
+    if (strpos($text, 'data-entity-type') !== FALSE && (strpos($text, 'data-entity-embed-display') !== FALSE || strpos($text, 'data-view-mode') !== FALSE)) {
       $dom = Html::load($text);
       $xpath = new \DOMXPath($dom);
 
-      foreach ($xpath->query('//*[@data-entity-type and @data-view-mode]') as $node) {
+      foreach ($xpath->query('//*[@data-entity-type and (@data-entity-uuid or @data-entity-id) and (@data-entity-embed-display or @data-view-mode)]') as $node) {
         $entity_type = $node->getAttribute('data-entity-type');
         $entity = NULL;
         $view_mode = $node->getAttribute('data-view-mode');
```
boztek: $view_mode not being used.

src/EntityEmbedDisplay/EntityEmbedDisplayInterface.php
davereid, line 20: Because eventually we should be able to replace setContextValue(), getContextValues(), and getContextValue() with ContextAwarePluginBase, but the D8 context system has bugs that prevents this from working.

All other methods are basically necessary to fulfill interfaces and not defined by any base classes.

src/EntityEmbedDisplay/FieldFormatterEntityEmbedDisplayBase.php
```diff
@@ -0,0 +1,120 @@
...
+
+    if ($langcode = $this->getAttributeValue('langcode')) {
+      $items->setLangcode($langcode);
+    }
+
+    $formatter = $this->getFormatter($definition);
+    // Prepare, expects an array of items, keyed by parent entity ID.
+    $formatter->prepareView(array($node->id() => $items));
+    $build = $formatter->viewElements($items);
+    show($build[0]);
```
boztek, line 72: Is this definitely needed? Having trouble understanding why it's required.

davereid, line 72: Yes, because whatever is inside viewElements() results in $build[0]['#printed'] being a 'TRUE' value, which means it will not get output unless we call show() on it.

entity_embed.api.php
davereid, line 29: I'll file this as a follow-up todo. This hook existed prior to this pull request anyway.

entity_embed.api.php
davereid, line 15: Will leave as a todo for now with a follow-up.

src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
stevector, line 57: Yes. I just spoke with Dave. He's planning on merging this PR shortly after a few more consultations. I'll start a new branch on top of this one for doing more injection throughout.

davereid:
@jibran Thanks for the reviews! The pattern I saw used in core, is that if it's a method defined by an interface, I can use {@inheritdoc}, which I think I have used but I found a few more that I could switch from "Implements...." to {@inheritdoc}

davereid:
Wee! Merged and complete! Thanks everyone for reviews and feedback!

slashrsm:
Yay!

slashrsm:
Just before ice cream ;)

PR #22, opened 2014-05-30
Basic CKEditor Integration

cs-shadow:
Right now integration is in very early stage. But it works as a proof of concept.

Details here: https://drupal.org/node/2277077


entity_embed.routing.yml
```diff
@@ -0,0 +1,7 @@
+entity_embed.ckeditor_form:
+  path: '/admin/entity-embed/embed'
+  defaults:
+    _form: '\Drupal\entity_embed\Form\EntityEmbedCkeditorForm'
+    _title: 'Embed an entity'
+  requirements:
+    _user_is_logged_in: 'TRUE'
```
davereid: I'm looking at editor.routing.yml and was expecting the path to be more like those which uses {filter_format} and proper access checking.

``` yaml

editor.image_dialog:
  path: '/editor/dialog/image/{filter_format}'
  defaults:
    _form: '\Drupal\editor\Form\EditorImageDialog'
    _title: 'Upload image'
  requirements:
    _entity_access: 'filter_format.use'
```

Should this routing path also depend if the editor module is enabled? It's not listed as a dependency currently for entity_embed.

src/Form/EntityEmbedCkeditorForm.php
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Provides a form to embed entities by specifying data attributes.
+ */
+class EntityEmbedCkeditorForm extends FormBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'entity_embed_ckedtor_form';
```
davereid: ckeditor <> ckeditor

src/Form/EntityEmbedCkeditorForm.php
```diff
@@ -0,0 +1,108 @@
...
+    $form['entity_type'] = array(
+      '#type' => 'select',
+      '#name' => 'entity_type',
+      '#title' => 'Entity type',
+      '#options' => array(
+        'node' => 'Node',
+        'others' => 'Others',
+      ),
+    );
+    $form['embed_method'] = array(
```
davereid: I'm not sure we need a UI element for this. We should be able to use [Uuid::isValid()](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Uuid%21Uuid.php/function/Uuid%3A%3AisValid/8) on the $form['entity'] value to detect if a UUID was provided or not.

src/Form/EntityEmbedCkeditorForm.php
```diff
@@ -0,0 +1,108 @@
...
+      '#type' => 'select',
+      '#name' => 'align',
+      '#title' => 'Align',
+      '#options' => array(
+        'none' => 'None',
+        'left' => 'Left',
+        'center' => 'Center',
+        'right' => 'Right',
+      ),
+    );
```
davereid: I think we need a 'Caption' checkbox as well here.

src/Form/EntityEmbedCkeditorForm.php
```diff
@@ -0,0 +1,108 @@
...
+use Drupal\Core\Ajax\AjaxResponse;
+use Drupal\Core\Ajax\CloseModalDialogCommand;
+use Drupal\Core\Ajax\HtmlCommand;
+use Drupal\Core\Form\FormBase;
+use Drupal\entity_embed\Ajax\EntityEmbedDialogSave;
+
+/**
+ * Provides a form to embed entities by specifying data attributes.
+ */
+class EntityEmbedCkeditorForm extends FormBase {
```
davereid: All of core files use class names with CKEditor not Ckeditor, so we should probably follow that standard.

entity_embed.routing.yml
cs-shadow: Yes, the editor module should also be a dependency (because its not required anywhere else.)

EDIT: Done.

js/plugins/entityembed/plugin.js
```diff
@@ -0,0 +1,81 @@
...
+      }
+    }
+
+  });
+
+  /**
+   * Function to save the data attributes specified in the modal.
+   */
+  Drupal.AjaxCommands.prototype.entityembedDialogSave = function (ajax, response, status) {
+    var editor = CKEDITOR.instances['edit-body-0-value'];
```
webflo: Please get rid of the hardcoded field name/ckeditor instance id.

js/plugins/entityembed/plugin.js
```diff
@@ -0,0 +1,93 @@
...
+        });
+      }
+    }
+
+  });
+
+  /**
+   * Function to save the data attributes specified in the modal.
+   */
+  Drupal.AjaxCommands.prototype.entityembedDialogSave = function (ajax, response, status) {
```
jibran, line 52: I think this should have the separate file.

PR #23, opened 2014-06-01
Embed form converted to two-step form.

cs-shadow:
File names have become little weird, so any suggestions are most welcome.


js/plugins/entityembed/plugin.js
```diff
@@ -47,9 +47,36 @@
   });
 
   /**
+   * Function to allow selection of entity to be embedded.
+   */
+  Drupal.AjaxCommands.prototype.entityembedSelectDialogSave = function (ajax, response, status) {
```
jibran: I think it is better if we can move AjaxCommands in to separate JS files. I also suggested that in #22.

cs-shadow: I'll just do that. Having some troubles with libraries.yml. As soon as that is fixed, I'll move the AjaxCommands to a separate file.

jibran: Thanks

js/plugins/entityembed/plugin.js
```diff
@@ -47,9 +47,36 @@
...
+    // Populate existing values, i.e. embed_method, entity_type and entity.
+    var existingValues = {};
+    existingValues['entity-type'] = response.values.entity_type;
+    existingValues['embed-method'] = response.values.embed_method;
+    existingValues['entity'] = response.values.entity;
+    existingValues['editor-id'] = editor.name;
+    var saveCallback = function(values) {
+    };
+    // Open the dialog for the entity embed form.
+    Drupal.ckeditor.openDialog(editor, Drupal.url('entity-embed/dialog/entity_embed/' + editor.config.drupal.format), existingValues, saveCallback, dialogSettings);
```
slashrsm: Think a little about how to make this steps configurable. Let's say that I want to completely swap out this step and replace it with something else (or even remove it entirely and just use some defaults). That should be possible and shouldn't allow much hacking with alters and stuff. Could we utilize plugins to achieve that?

This can be done as part of a follow-up issue.

js/plugins/entityembed/plugin.js
cs-shadow: AjaxCommands moved to separate JS file.

PR #24, opened 2014-06-03
Fixing error with compatibility with Interface

stevector:
I installed this module and got an error. Here is a fix.

```
Fatal error: Declaration of Drupal\entity_embed\Plugin\Filter\EntityEmbedFilter::process() must be compatible with Drupal\filter\Plugin\FilterInterface::process($text, $langcode, $cache, $cache_id) in /Users/persch/Sites/d8_dev/docroot/modules/entity_embed/src/Plugin/Filter/EntityEmbedFilter.php on line 34
```

screen shot 2014-06-03 at 4 03 27 pm


cs-shadow:
For this module to work properly, Drupal 8 core must have a current patch from https://drupal.org/node/2217877 applied.

Its mentioned in the requirements of this module: https://github.com/drupal-media/entity_embed#requirements

Hopefully, the core issue will get committed soon and this requirement will no longer be required.

PR #25, opened 2014-06-04
Validate form for presence of required fields

cs-shadow:
Make following fields compulsory: entity_type, entity and view mode.
If any error occurs, show the error in the dialog before proceeding further.

At the final step before submit happens, validate form to ensure all the required fields are present (from previous step as well as this step)

Validating presence of fields from previous step just to be sure, a conservative step.

d.o issue: https://drupal.org/node/2279175

### DIFF
--- a/src/Form/EntityEmbedCKEditorEmbedForm.php
+++ b/src/Form/EntityEmbedCKEditorEmbedForm.php
@@ -113,7 +113,7 @@ class EntityEmbedCKEditorEmbedForm extends FormBase {
       $status_messages = array('#theme' => 'status_messages');
       $output = drupal_render($form);
       $output = '<div>' . drupal_render($status_messages) . $output . '</div>';
-      $response->addCommand(new HtmlCommand('#entity-embed-ckeditor-select-form', $output));
+      $response->addCommand(new HtmlCommand('#entity-embed-ckeditor-embed-form', $output));
     }
     else {
       $response->addCommand(new EntityEmbedSubmitDialogSave($form_state['values']));
--- a/src/Form/EntityEmbedCKEditorEmbedForm.php
+++ b/src/Form/EntityEmbedCKEditorEmbedForm.php
@@ -56,6 +56,7 @@ class EntityEmbedCKEditorEmbedForm extends FormBase {
       '#name' => 'view_mode',
       '#title' => 'View Mode',
       '#options' => $view_modes,
+      '#required' => TRUE,
     );
     $form['display_links'] = array(
       '#type' => 'checkbox',
--- a/src/Form/EntityEmbedCKEditorEmbedForm.php
+++ b/src/Form/EntityEmbedCKEditorEmbedForm.php
@@ -107,8 +107,18 @@ class EntityEmbedCKEditorEmbedForm extends FormBase {
   public function submitForm(array &$form, array &$form_state) {
     $response = new AjaxResponse();
 
-    $response->addCommand(new EntityEmbedSubmitDialogSave($form_state['values']));
-    $response->addCommand(new CloseModalDialogCommand());
+    // Display errors in form, if any.
+    if (form_get_errors($form_state)) {
+      unset($form['#prefix'], $form['#suffix']);
+      $status_messages = array('#theme' => 'status_messages');
+      $output = drupal_render($form);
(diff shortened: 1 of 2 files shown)


src/Form/EntityEmbedCKEditorEmbedForm.php
```diff
@@ -107,8 +108,18 @@ public function buildForm(array $form, array &$form_state) {
...
 
-    $response->addCommand(new EntityEmbedSubmitDialogSave($form_state['values']));
-    $response->addCommand(new CloseModalDialogCommand());
+    // Display errors in form, if any.
+    if (form_get_errors($form_state)) {
+      unset($form['#prefix'], $form['#suffix']);
+      $status_messages = array('#theme' => 'status_messages');
+      $output = drupal_render($form);
+      $output = '<div>' . drupal_render($status_messages) . $output . '</div>';
+      $response->addCommand(new HtmlCommand('#entity-embed-ckeditor-embed-form', $output));
```
slashrsm: This shouldn't be hardcoded.

src/Form/EntityEmbedCKEditorSelectForm.php
```diff
@@ -74,19 +75,29 @@ public function buildForm(array $form, array &$form_state) {
...
-    $entity = $values['entity'];
-    if (Uuid::isValid($entity)) {
-      $values['embed_method'] = 'uuid';
+    // Display errors in form, if any.
+    if (form_get_errors($form_state)) {
+      unset($form['#prefix'], $form['#suffix']);
+      $status_messages = array('#theme' => 'status_messages');
+      $output = drupal_render($form);
+      $output = '<div>' . drupal_render($status_messages) . $output . '</div>';
+      $response->addCommand(new HtmlCommand('#entity-embed-ckeditor-select-form', $output));
```
slashrsm: Same as above.

src/Form/EntityEmbedCKEditorEmbedForm.php
cs-shadow: This is same as how the core does it: https://api.drupal.org/api/drupal/core%21modules%21editor%21src%21Form%21EditorImageDialog.php/function/EditorImageDialog%3A%3AsubmitForm/8
and it seems fine because the form id itself is being returned by this file only.

Another way to do this will to be to pass around the form id as a hidden field, but finding out the form id itself will be a bit hacky solution.

slashrsm: FormID should be availabe in $form and/or $form_state.

src/Form/EntityEmbedCKEditorSelectForm.php
```diff
@@ -74,19 +75,29 @@ public function buildForm(array $form, array &$form_state) {
...
     $response = new AjaxResponse();
 
-    // Detect if a valid UUID was specified. Set embed method based based on
-    // whether or not it is a valid UUID.
-    $values = $form_state['values'];
-    $entity = $values['entity'];
-    if (Uuid::isValid($entity)) {
-      $values['embed_method'] = 'uuid';
+    // Display errors in form, if any.
+    if (form_get_errors($form_state)) {
```
boztek: Deprecated - use \Drupal::formBuilder()->getErrors($form_state)

src/Form/EntityEmbedCKEditorEmbedForm.php
```diff
@@ -104,11 +105,34 @@ public function buildForm(array $form, array &$form_state) {
...
+  /**
+   * {@inheritdoc}
+   */
   public function submitForm(array &$form, array &$form_state) {
     $response = new AjaxResponse();
 
-    $response->addCommand(new EntityEmbedSubmitDialogSave($form_state['values']));
-    $response->addCommand(new CloseModalDialogCommand());
+    // Display errors in form, if any.
+    if (form_get_errors($form_state)) {
```
boztek: Deprecated - use \Drupal::formBuilder()->getErrors($form_state)

src/Form/EntityEmbedCKEditorSelectForm.php
cs-shadow: Thanks for this catch

PR #27, opened 2014-06-04
Add a Back button in the second step of the Embed form

cs-shadow:
d.o issue: https://drupal.org/node/2279269

### DIFF
--- a/js/entity_embed.ajax.js
+++ b/js/entity_embed.ajax.js
@@ -27,10 +27,8 @@
     existingValues['embed-method'] = response.values.embed_method;
     existingValues['entity'] = response.values.entity;
     existingValues['editor-id'] = editor.name;
-    var saveCallback = function(values) {
-    };
     // Open the dialog for the entity embed form.
-    Drupal.ckeditor.openDialog(editor, Drupal.url('entity-embed/dialog/entity_embed/' + editor.config.drupal.format), existingValues, saveCallback, dialogSettings);
+    Drupal.ckeditor.openDialog(editor, Drupal.url('entity-embed/dialog/entity_embed/' + editor.config.drupal.format), existingValues, null, dialogSettings);
   };
 
   /**
@@ -52,10 +50,8 @@
     existingValues['entity-type'] = response.values.entity_type;
     existingValues['entity'] = response.values.entity;
     existingValues['editor-id'] = editor.name;
-    var saveCallback = function(values) {
-    };
     // Open the dialog for the entity embed form.
-    Drupal.ckeditor.openDialog(editor, Drupal.url('entity-embed/dialog/entity_select/' + editor.config.drupal.format), existingValues, saveCallback, dialogSettings);
+    Drupal.ckeditor.openDialog(editor, Drupal.url('entity-embed/dialog/entity_select/' + editor.config.drupal.format), existingValues, null, dialogSettings);
   };
 
   /**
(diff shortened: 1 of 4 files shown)


js/entity_embed.ajax.js
```diff
@@ -34,6 +34,31 @@
...
+      resizable: false,
+      minWidth: 800
+    };
+
+    // Populate existing values, i.e. embed_method, entity_type and entity.
+    var existingValues = {};
+    existingValues['entity-type'] = response.values.entity_type;
+    existingValues['entity'] = response.values.entity;
+    existingValues['editor-id'] = editor.name;
+    var saveCallback = function(values) {
```
slashrsm: Do we need this since it is empty? Could we throw empty function declaration directly inside the openDialog() if no? Also, "values" is undefined var.

src/Form/EntityEmbedCKEditorEmbedForm.php
```diff
@@ -79,6 +80,16 @@ public function buildForm(array $form, array &$form_state) {
...
+      // No regular submit-handler. This form only works via JavaScript.
+      '#submit' => array(),
+      '#ajax' => array(
+        'callback' => array($this, 'goBack'),
+        'event' => 'click',
+      ),
+    );
     $form['actions']['save_modal'] = array(
       '#type' => 'submit',
       '#value' => 'Save',
```
jibran, line 95: No `t` here as well.

js/entity_embed.ajax.js
cs-shadow: We can put the empty function declaration directly inside the openDialog(), but it will not look very neat then. It will look like:

``` html
Drupal.ckeditor.openDialog(editor, Drupal.url('entity-embed/dialog/entity_select/' + editor.config.drupal.format), existingValues={}, saveCallback=function(values){}, dialogSettings);
```

Also, values in not undefined. Its a parameter to the function.
IMO, changing this will make the function call look ugly. Let me know if you still want to change this.

src/Form/EntityEmbedCKEditorEmbedForm.php
cs-shadow, line 95: Thanks. Wrapped value in `t`

src/Form/EntityEmbedCKEditorEmbedForm.php
```diff
@@ -79,6 +80,16 @@ public function buildForm(array $form, array &$form_state) {
       '#title' => 'Show Caption',
     );
     $form['actions'] = array('#type' => 'actions');
+    $form['actions']['go_back'] = array(
+      '#type' => 'submit',
+      '#value' => '$this->t(Back'),
```
boztek: This doesn't look right?

```
$this->t('Back'),
```

cs-shadow: What looks wrong (syntactically or logically)?

(This basically adds a button to go back to previous step of the form)

boztek: Looks like you have

```
'#value' => '$this->t(Back'),
```

so some misplaced quote marks?

cs-shadow: Sorry, My bad

boztek:
This is not currently working in the situation where you change the entity ID after going back. It is still saving the entity ID first entered.

Tested against latest form-validation and 8.x-1.x branches and other than that it merges clean and seems to work.

cs-shadow:
Its working fine now. Fixed the issue where the entity ID was not being changed after going back.

PR #28, opened 2014-06-05
Update EntityEmbedFilter::process with FilterInterface change

boztek:


slashrsm:
You need core patch to run this filter: https://drupal.org/node/2217877

Patch changes signature of process() function.

PR #30, opened 2014-06-06
Embed display review

boztek:
Includes comments from @jibran
Doesn't include the DI suggestion yet - https://github.com/drupal-media/entity_embed/pull/21#discussion_r13459303


davereid:
Eek, I think I already have some of these fixed in the embed-display plugin branch. Let's see if this is still mergeable after I merge #21. Also note that @stevector is looking to work on the DI fixes (https://drupal.org/node/2281487 and https://drupal.org/node/2281491).

PR #32, opened 2014-06-06
[WIP] Adding more dependency injection.

stevector:
Here's the d.o issue https://drupal.org/node/2281491

As of the creation of this pull request there is only one commit that adds a create method. It isn't adding any benefit yet and it is a work in progress.

### DIFF
--- a/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
@@ -23,6 +23,13 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 class EntityEmbedDefaultDisplay extends EntityEmbedDisplayBase implements ContainerFactoryPluginInterface {
 
   /**
+   * The entity manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
+
+  /**
    * {@inheritdoc}
    *
    * @param EntityManagerInterface $entity_manager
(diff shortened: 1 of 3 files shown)


davereid:
So just an initial question - the plugin base doesn't actually use the entity manager anywhere. Should the individual plugins be responsible for pulling stuff from the container as they need them, or is the plugin base responsible? I have a feeling that the individual plugins should be responsible.

stevector:
I really don't know. my thinking was "other child classes beside the default will need the entityManager" I'll look for a pattern elsewhere in core.

stevector:
`ActionAddForm` makes me think you are correct. So does `CustomBlockController`. I'll switch it up.

slashrsm:
I'd say individual plugins are responsible for things they need.

stevector:
Thanks @slashrsm. I've pushed the refactor.

src/EntityEmbedDisplay/EntityEmbedDefaultDisplay.php
```diff
@@ -17,7 +20,30 @@
  *
  * @todo Should this use a derivative? http://cgit.drupalcode.org/page_manager/tree/src/Plugin/Deriver/EntityViewDeriver.php
  */
-class EntityEmbedDefaultDisplay extends EntityEmbedDisplayBase {
+class EntityEmbedDefaultDisplay extends EntityEmbedDisplayBase implements ContainerFactoryPluginInterface {
+
```
davereid, line 24: I think we're missing a declaration for the $this->entityManager class property.

Copied from EntityEmbedFilter:

``` php
  /**
   * The entity manager service.
   *
   * @var \Drupal\Core\Entity\EntityManagerInterface
   */
  protected $entityManager;
```

stevector, line 24: Thanks! I just added it.

davereid:
Great, looks good!

PR #33, opened 2014-06-06
Raise an error if specified entity_type or entity does not exists

cs-shadow:


jibran:
I think we should wait for #34. It'll make it a lot simpler.

PR #34, opened 2014-06-07
Add EntityHelperTrait.

davereid:
This helps reduce the amount of code we're having to use for entity loading and rendering, while also remaining testable. We could easily be using this on the CKEditor integration branch too.

### DIFF
--- a/src/Plugin/Filter/EntityEmbedFilter.php
+++ b/src/Plugin/Filter/EntityEmbedFilter.php
@@ -12,6 +12,7 @@ use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityManagerInterface;
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\entity_embed\EntityHelperTrait;
 use Drupal\entity_embed\RecursiveRenderingException;
 use Drupal\filter\FilterProcessResult;
 use Drupal\filter\Plugin\FilterBase;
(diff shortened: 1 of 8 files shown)


src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -98,18 +99,11 @@ public function process($text, $langcode) {
...
-
-            $entity_type_definition = $this->entityManager->getDefinition($entity_type);
-            $uuid_key = $entity_type_definition->getKey('uuid');
-            $controller = $this->entityManager->getStorage($entity_type);
-            $entities = $controller->loadByProperties(array($uuid_key => $uuid));
-            $entity = reset($entities);
+            $entity = $this->entityManager->load($entity_type, $uuid);
           }
           elseif ($node->hasAttribute('data-entity-id')) {
             $id = $node->getAttribute('data-entity-id');
```
jibran: so why are we using two attributes then if both are same for entity manger then why not just use one `data-entity-id` and remove `data-entity-uuid`.

davereid: Because we want to encourage use of uuid by default. and because we have special logic in the id version to add the uuid back into the DOM as an attribute on filtering.

cs-shadow: The reasoning is fine with me, but there's something else that's bugging me. Once in WYSIWYG we determine whether the specified ID is a valid UUID or not and fill in 'data-entity-id' or 'data-entity-uuid' accordingly. Then again here in the load function we are checking whether we have a valid UUID or not. Can we do something to reduce this to just once check rather than doing the same check twice?

davereid:
I switched to using a trait since that actually seemed to fit the better use case, rather than having to make sure our class implemented _all_ of the methods of the EntityManagerInterface.

davereid:
I've tested this extensively and gives us a nice place to gather all these helpers that we need that work with entities. Merging!

PR #38, opened 2014-06-10
Adds EntityEmbedPostRenderCache service class.

davereid:
Depends on the core patch from https://drupal.org/project/2247779.


README.md
```diff
@@ -2,8 +2,8 @@
 
 ## Requirements
 
-* Drupal 8 core must have a current patch from https://drupal.org/node/2217877
-  applied.
+* Drupal 8 core must have a current patch from
+  https://drupal.org/project/2247779 applied.
```
jibran: should be https://drupal.org/node/2247779 :)

davereid: Thanks!

## 

Dave Reid
dave@davereid.net | http://www.davereid.net/ | 402-431-2412

davereid:
We can hold off on this for now I think, but like the previous core patch, we'll have to be using this anyway at some point, so we might as well merge it sometime in the near future as well.

PR #40, opened 2014-06-12
Add back button in the second step of the form

cs-shadow:
d.o issue: https://drupal.org/node/2279269

### DIFF
--- a/src/Form/EntityEmbedDialog.php
+++ b/src/Form/EntityEmbedDialog.php
@@ -177,6 +177,16 @@ class EntityEmbedDialog extends FormBase {
         $form['actions'] = array(
           '#type' => 'actions',
         );
+        $form['actions']['back'] = array(
+          '#type' => 'submit',
+          '#value' => $this->t('Back'),
+          // No regular submit-handler. This form only works via JavaScript.
+          '#submit' => array(),
+          '#ajax' => array(
+            'callback' => array($this, 'goBack'),
+            'event' => 'click',
+          ),
+        );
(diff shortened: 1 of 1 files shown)


src/Form/EntityEmbedDialog.php
```diff
@@ -223,4 +233,27 @@ public function submitForm(array &$form, array &$form_state) {
...
+   *   An associative array containing the structure of the form.
+   * @param array $form_state
+   *   An associative array containing the current state of the form.
+   */
+  public function goBack(array &$form, array &$form_state) {
+    $response = new AjaxResponse();
+
+    $form_state['rebuild'] = TRUE;
+    $form_state['step'] = 'select';
+    $rebuild_form = \Drupal::formBuilder()->rebuildForm('entity_embed_dialog', $form_state, $form);
```
slashrsm, line 307: Is it possible to inject this?

slashrsm:
It looks that this branch needs re-roll.

src/Form/EntityEmbedDialog.php
cs-shadow, line 307: formBuilder() is used in several places throughout the code. If we are going to try to inject this, I'd suggest do that as part of a separate issue, because this needs to be changed at other places also.

slashrsm, line 307: OK. Agree. Please open follow-up.

cs-shadow, line 307: Follow-up issue: https://drupal.org/node/2287553

PR #41, opened 2014-06-12
Fix coding standards

cs-shadow:


src/EntityHelperTrait.php
```diff
@@ -26,7 +27,7 @@
    *
    * @var \Drupal\Core\Entity\EntityManagerInterface
    */
-  protected $entityManager;
+  protected $entity_manager;
```
davereid: Not sure this is actually necessary?

src/Annotation/EntityEmbedDisplay.php
```diff
@@ -42,6 +42,6 @@ class EntityEmbedDisplay extends Plugin {
    *
    * @var bool|array
    */
-  public $entity_types = FALSE;
+  public $entityTypes = FALSE;
```
davereid: The key used in the annotation is actually 'entity_types' so we can't just change this without also changing the annotations. I think we'll probably want to leave this one alone.

src/EntityHelperTrait.php
cs-shadow: `29 | ERROR   | Variable "entityManager" is camel caps format. do not use 
        |                 | mixed case (camelCase), use lower case and _`

This is code-sniffer output, so I thought it might be a good idea to change to lower case

davereid: I thnk it's fine to ignore. Core definitely uses way more camel caps format for variables now with D8, so I think it's more of a "code-sniffer would need to be updated for D8 standards" situation.

cs-shadow: Oh, okay. Reverted this change.

src/Plugin/Derivative/FieldFormatterDeriver.php
```diff
@@ -24,16 +24,16 @@ class FieldFormatterDeriver extends DerivativeBase implements ContainerDerivativ
    *
    * @var \Drupal\Core\Field\FormatterPluginManager.
    */
-  protected $formatterManager;
+  protected $formatter_manager;
```
davereid: Let's revert this change as well.

PR #42, opened 2014-06-12
Update EntityEmbedDisplayManager no longer needs LanguageManager.

davereid:
https://drupal.org/node/2281905 was committed to core, so we need to update our plugin manager since we extend the DefaultPluginManager from core.


davereid:
Tested and confirmed this works.

PR #43, opened 2014-06-12
Remove patch from core issue as requiremnt since the patch is committed

cs-shadow:


README.md
```diff
@@ -1,10 +1,5 @@
 # Entity Embed Module
 
-## Requirements
-
-* Drupal 8 core must have a current patch from https://drupal.org/node/2217877
```
davereid, line 5: Should we say instead something like "An up-to-date 8.x version of Drupal core" or something to be clear this module will not work with the last "alpha" release?

davereid:
Also cross-referencing https://github.com/drupal-media/entity_embed/pull/38 since that re-adds another core patch as dependency.

README.md
cs-shadow, line 5: How does the new requirement sounds?

PR #44, opened 2014-06-15
Add Display plugin options

cs-shadow:
d.o issue: https://drupal.org/node/2283691

When this issue goes through, following follow-up issues will be needed:
- Remove redundant 'view-mode' field from the form.
- Use rebuildEmbedForm() in submitForm() when form['state'] is select.

### DIFF
--- a/src/Form/EntityEmbedDialog.php
+++ b/src/Form/EntityEmbedDialog.php
@@ -157,11 +157,14 @@ class EntityEmbedDialog extends FormBase {
           '#title' => $this->t('Display as'),
           '#options' => $manager->getDefinitionOptionsForEntity($entity),
           '#ajax' => array(
-            'callback' => array($this, 'rebuildEmbedForm'),
+            'callback' => array($this, 'updatePluginConfigurationForm'),
             'event' => 'change',
+            'wrapper' => 'data-entity-embed-settings-form',
           ),
         );
         $form['attributes']['data-entity-embed-settings'] = $display->buildConfigurationForm($form, $form_state);
+        $form['attributes']['data-entity-embed-settings']['#prefix'] = '<div id="data-entity-embed-settings-form">';
+        $form['attributes']['data-entity-embed-settings']['#suffix'] = '</div>';
         $form['attributes']['data-view-mode'] = array(
           '#type' => 'select',
           '#title' => $this->t('View Mode'),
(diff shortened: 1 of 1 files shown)


src/Form/EntityEmbedDialog.php
```diff
@@ -63,6 +65,8 @@ public function buildForm(array $form, array &$form_state) {
     $form['#suffix'] = '</div>';
     $form['#attached']['library'][] = 'entity_embed/entity_embed.ajax';
 
+    $manager = \Drupal::service('plugin.manager.entity_embed.display');
```
slashrsm: Inject?

src/Form/EntityEmbedDialog.php
```diff
@@ -221,4 +252,19 @@ public function submitForm(array &$form, array &$form_state) {
     return $response;
   }
 
+  public function rebuildEmbedForm(array &$form, array &$form_state) {
```
slashrsm: Missing phpdoc block.

src/Form/EntityEmbedDialog.php
```diff
@@ -221,4 +273,27 @@ public function submitForm(array &$form, array &$form_state) {
...
 
+  /**
+   * Form submission handler to build or rebuild the embed step of the form.
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param array $form_state
+   *   An associative array containing the current state of the form.
+   */
+  public function rebuildEmbedForm(array &$form, array &$form_state) {
```
davereid: I think this could be massively simplified by using the actual AJAX framework to only replace part of the form that needs to be updated (the `$form['attributes']['data-entity-embed-settings']` element). See https://api.drupal.org/api/drupal/core!includes!ajax.inc/group/ajax/8

src/Form/EntityEmbedDialog.php
```diff
@@ -212,6 +257,13 @@ public function submitForm(array &$form, array &$form_state) {
           break;
 
         case 'embed':
+          // Serialize entity embed settings to JSON string.
+          $entity_embed_settings = JSON::encode($form_state['values']['attributes']['entity-embed-settings']);
+          $entity_embed_settings = str_replace('"', "'", $entity_embed_settings);
```
davereid: I'm not sure this is necessary? In my quick testing, I think when saved it should escape double quotes to a `&quot;` character in the JSON? Or is that not happening?

src/Form/EntityEmbedDialog.php
```diff
@@ -104,6 +129,16 @@ public function buildForm(array $form, array &$form_state) {
...
       case 'embed':
         $entity = $this->loadEntity($entity_element['data-entity-type'], $entity_element['data-entity-uuid'] ?: $entity_element['data-entity-id']);
+
+        $plugin_configuration = $entity_element;
+        $plugin_configuration['entity'] = $entity;
+        $plugin_id = $entity_element['data-entity-embed-display'];
+        $display = $manager->createInstance($plugin_id);
+        // Set attributes as context values.
+        foreach ($plugin_configuration as $name => $value) {
+          $display->setContextValue($name, $value);
```
davereid: Only the 'entity' value actually needs to be passed in using setContextValue(). The rest should be passed in using setAttributes($array).

src/Form/EntityEmbedDialog.php
```diff
@@ -121,6 +156,16 @@ public function buildForm(array $form, array &$form_state) {
           '#type' => 'value',
           '#value' => $entity_element['data-entity-uuid'],
         );
+        $form['attributes']['data-entity-embed-display'] = array(
+          '#type' => 'select',
+          '#title' => $this->t('Display as'),
+          '#options' => $manager->getDefinitionOptionsForEntity($entity),
```
davereid, line 158: I wonder if we should add some protection for what happens if #options is empty here.

src/Form/EntityEmbedDialog.php
cs-shadow: I thought double escaping might cause problems while decoding.
Reverted that change now.

src/Form/EntityEmbedDialog.php
cs-shadow, line 158: Maybe we can just show the list of view-modes and remove this field.
But I think that should be done as part a follow-up issue after we decide what to do with the view-mode field.

PR #45, opened 2014-06-15
Use entity-embed-settings instead of data-entity-embed-settings

cs-shadow:
In code, the data attribute is 'entity-embed-settings' while in readme the attribute is mentioned as 'data-entity-embed-settings', so fixing the attribute in readme to match the actual attribute.


slashrsm:
It is converted to "entity-embed-settings" in $context array, but still expects data- prefix when looking for attributes. See https://github.com/drupal-media/entity_embed/blob/8.x-1.x/src/Plugin/Filter/EntityEmbedFilter.php#L103

cs-shadow:
Ah, my bad. In that case, this PR can be declined.

PR #47, opened 2014-06-15
Fix coding standard

cs-shadow:
Nitpick: According to the standards core follows, there shouldn't be a full-stop preceding the filter name.


src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -22,7 +22,7 @@
  *
  * @Filter(
  *   id = "entity_embed",
- *   title = @Translation("Display embedded entities."),
```
slashrsm, line 25: Hm... Are you sure? AFAIK titles end w/o a period.

slashrsm, line 25: Ah.... Forget it...

PR #48, opened 2014-06-15
Typo in Exception name.

jibran:
Fatal error: Class 'Drupal\entity_embed\RecursiveRenderingException' not found in /var/www/d8/modules/entity_embed/src/EntityHelperTrait.php on line 154 thanks to CS


jibran:
Thanks to @cs-shadow for the fix.

cs-shadow:
Funny that I also created the same PR at #49 just a few moments later. But that can be closed now.

PR #49, opened 2014-06-15
Fix typo in filename

cs-shadow:
There's a typo in name of the file: ResuriveRenderingException.php, it should be RecursiveRenderingException.php.
Surprising, this never got noticed.


cs-shadow:
Closed, since #48 fixes this.

PR #51, opened 2014-06-17
Add module to Media package

cs-shadow:


davereid:
Hrm, I'm not set on using a 'Media' package in D8 for modules. It's already kind of against the guidelines for package information. Plus the module is so non-media specific. I would be inclined to list it in a 'Filter' section instead since that's what the primary purpose is.

cs-shadow:
Sorry, my bad for the commit. It was a mistake, had to push another branch this got in somehow. Reverted the commit.

cs-shadow:
Entity Browser is currently in Media package, so I thought that'd be the right place for it. Though 'Filter' package fits better. Lets put this in Filter package if @slashrsm agrees.

davereid:
I'd argue that Entity Browser also isn't a good fit for the Media package as well - it's a generic API module that is used by Media related modules. I think the package should be reserved for things inside https://drupal.org/project/media and related modules (like media_youtube, etc).

slashrsm:
Agree. Filter sounds much better. Also agree that browser does not belong into Media package. We will change that. I am planning to dedicate entire Friday to browser.

cs-shadow:
New PR to add it Filter package: #54

PR #52, opened 2014-06-17
Add a preview for the selected entity in WYSIWYG

cs-shadow:
d.o issue: https://drupal.org/node/2280547

### DIFF
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -47,9 +47,6 @@
         // Generate the preview of the element and render it.
         upcast: function (element) {
           var attributes = element.attributes;
-          if (element.name != 'div') {
-            return;
-          }
           if (attributes['data-entity-type'] === undefined || (attributes['data-entity-id'] === undefined && attributes['data-entity-uuid'] === undefined)) {
             return;
           }
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -51,7 +51,6 @@
             return;
           }
           var request = {};
-          request['filter-format'] = editor.config.drupal.format;
           request['value'] = element.getOuterHtml();
           jQuery.ajax({
             url: Drupal.url('entity-embed/preview/' + editor.config.drupal.format + '?' + jQuery.param(request)),
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -53,7 +53,14 @@
           if (attributes['data-entity-type'] === undefined || (attributes['data-entity-id'] === undefined && attributes['data-entity-uuid'] === undefined)) {
             return;
           }
-          element.setHtml('Set preview here.');
+          jQuery.ajax({
+            url: Drupal.url('entity-embed/preview/?' + jQuery.param(attributes)),
+            dataType: 'json',
+            async: false,
+            success: function (data) {
+              element.setHtml(data.content);
+            }
+          });
           return element;
         },
 
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -50,8 +50,11 @@
           if (attributes['data-entity-type'] === undefined || (attributes['data-entity-id'] === undefined && attributes['data-entity-uuid'] === undefined)) {
             return;
           }
+          var request = {};
+          request['filter-format'] = editor.config.drupal.format;
+          request['value'] = element.getOuterHtml();
           jQuery.ajax({
-            url: Drupal.url('entity-embed/preview/' + editor.config.drupal.format + '?' + jQuery.param(attributes)),
+            url: Drupal.url('entity-embed/preview/' + editor.config.drupal.format + '?' + jQuery.param(request)),
             dataType: 'json',
             async: false,
             success: function (data) {
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -13,7 +13,7 @@
 
     // Register the icon used for the toolbar button. It must be the same as the
     // name of the widget.
-    icons: 'entityembed',
+    icons: 'entity_embed',
 
     // The plugin initialization logic goes inside this method.
     init: function (editor) {
@@ -37,7 +37,7 @@
       });
 
       // Register the entity embed widget.
-      editor.widgets.add('entityembed', {
+      editor.widgets.add('entity_embed', {
         // Minimum HTML which is required by this widget to work.
         requiredContent: 'div[data-entity-type,data-entity-uuid]',
 
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -68,7 +68,8 @@
         // div element will be discarded by the CKEditor.
         // Using entity_type:entity_id as placeholder.
         downcast: function (element) {
-          element.setHtml('Set placeholder here.');
+          var attributes = element.attributes;
+          element.setHtml(attributes['data-entity-type'] + ':' + attributes['data-entity-id']);
           return element;
         }
       });
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -51,7 +51,7 @@
           if (element.name != 'div') {
             return;
           }
-          if (attributes['data-entity-type'] === undefined || (attributes['data-entity-type'] === undefined && attributes['data-entity-type'] === undefined)) {
+          if (attributes['data-entity-type'] === undefined || (attributes['data-entity-id'] === undefined && attributes['data-entity-uuid'] === undefined)) {
             return;
           }
           var attributes = element.attributes;
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -54,7 +54,7 @@
             return;
           }
           jQuery.ajax({
-            url: Drupal.url('entity-embed/preview/?' + jQuery.param(attributes)),
+            url: Drupal.url('entity-embed/preview/' + editor.config.drupal.format + '?' + jQuery.param(attributes)),
             dataType: 'json',
             async: false,
             success: function (data) {
--- a/js/plugins/entityembed/plugin.js
+++ b/js/plugins/entityembed/plugin.js
@@ -8,6 +8,9 @@
   "use strict";
 
   CKEDITOR.plugins.add('entityembed', {
+    // This plugin requires the Widgets System defined in the 'widget' plugin.
+    requires: 'widget',
+
     // Register the icon used for the toolbar button. It must be the same as the
     // name of the widget.
     icons: 'entityembed',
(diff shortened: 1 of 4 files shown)


js/plugins/entityembed/plugin.js
```diff
@@ -33,6 +36,44 @@
...
+        // Minimum HTML which is required by this widget to work.
+        requiredContent: 'div[data-entity-type,data-entity-uuid]',
+
+        // Add entity embed button
+        button: 'Insert entity',
+
+        // Generate the preview of the element and render it.
+        upcast: function (element) {
+          var attributes = element.attributes;
+          if (element.name != 'div') {
```
slashrsm: Is there any technical reason for this? It seems perfectly valid case that someone would use some other HTML element instead of a div.

js/plugins/entityembed/plugin.js
```diff
@@ -33,6 +36,44 @@
...
+            success: function (data) {
+              element.setHtml(data.content);
+            }
+          });
+          return element;
+        },
+
+        // Downcast the element. Do not set the html to be empty, otherwise the
+        // div element will be discarded by the CKEditor.
+        // Using entity_type:entity_id as placeholder.
```
slashrsm, line 61: It would be great if we could display exactly the same html as it is being saved into DB (i.e. div with attributes).

src/EntityEmbedController.php
```diff
@@ -0,0 +1,47 @@
...
+class EntityEmbedController extends ControllerBase {
+  use EntityHelperTrait;
+
+  /**
+   * Returns an Ajax response to generate preview of an entity.
+   *
+   * Expects the data attributes to be provided as GET parameters.
+   */
+  public function preview() {
+    $context = \Drupal::request()->query->all();
```
slashrsm: It is possible to inject this?

src/EntityEmbedController.php
```diff
@@ -0,0 +1,47 @@
...
+    $entity_embed_settings = Json::decode($context['data-entity-embed-settings']);
+    $entity_output = 'No preview available.';
+    try {
+      if ($entity = $this->loadEntity($context['data-entity-type'], $context['data-entity-id'])) {
```
slashrsm: UUID?

js/plugins/entityembed/plugin.js
cs-shadow: No technical reasons. We should all elements we'll support. Otherwise things may get messed up, say someone tries to insert an entity within an `` tag.

js/plugins/entityembed/plugin.js
cs-shadow, line 61: It is that only. This placeholder is just the inner HTML since innerHTML can't be empty. Outer HTML is exactly same as the user had entered. For example, `<div data-entity-id="1" data-view-mode="teaser" data-entity-type="node">Placeholder</div>`

src/EntityEmbedController.php
cs-shadow: Sure, will change this to UUID.

js/plugins/entityembed/plugin.js
slashrsm: I think we should fight this by making clear how to use this module. There are other things that this module allows and are very messy and wrong. 

Imagine body field + entity_embed as a layout manager for example. That is perfectly possible to implement but not a smart thing to do.

js/plugins/entityembed/plugin.js
slashrsm, line 61: Ah.... gothca. What about &nbsp;?

src/EntityEmbedController.php
slashrsm: You send both over so you should probably handle both cases (making UUID the first choice).

js/plugins/entityembed/plugin.js
cs-shadow: Actually meant what if someone uses the data attributes inside an `` tag, the text filter here removed that tag. I think we should restrict elements to just div, p and span. I don't see any need for other elements to be used.

slashrsm: <article>, <section>, <figure>, ...? You name it ;)

src/EntityEmbedController.php
cs-shadow: When I wrote this, I was assuming that preview will be generated only when someone uses our plugin to embed an entity and in which case both entity-id and entity-uuid will refer to same entity. But someone may also try to embed directly using source and in that case that assumption will not be valid. So, I'll add both cases here now. Good catch.

src/EntityEmbedController.php
cs-shadow: Yes. will just do that

js/plugins/entityembed/plugin.js
cs-shadow, line 61: The `&nbsp;` in your comment above got filtered out and I was wondering what good a question mark will do. Let me check if that works.

js/plugins/entityembed/plugin.js
cs-shadow: I see, Will remove that check then.

(@slashrsm, use backtick to wrap html, otherwise its just a bunch commas and fullstops.)

js/plugins/entityembed/plugin.js
cs-shadow, line 61: For some reason, `&nbsp;`doesn't seems to work here. We can decide on the placeholder later since php also sets placeholder in one place. We'll have make all of those follow same standards. We can do that in another issue.

davereid:
I'm just thinking here, rather than having to duplicate all the logic that the filter itself does in the preview controller, can we pass in the actual HTML element itself and run this?

``` php
  public function filterText(Request $request, FilterFormatInterface $filter_format) {
    $value = $request->get('value');
    if (!isset($value)) {
      throw new NotFoundHttpException();
    }

    $output = check_markup($value, $filter_format->id());
    return new JsonResponse($output);
  }
```

cs-shadow:
@daveried, as you suggested the preview function now expects the HTML element and filter format as GET params. Now check_markup() is used to generate preview instead of duplicating the logic of the filter.

slashrsm:
I've been playing with this and noticed some problems. Here is what I did:
1. Configured Full HTML format to use E button and enabled filter
2. Created node that is going to be embeddded
3. Started creating parent node
4. Selected Full HTML format
5. Got this error:

"NetworkError: 403 Forbidden - http://localhost/drupal8/entity-embed/dialog/entity-embed/full_html"
full_html
AjaxError: An AJAX HTTP error occurred. HTTP Result Code: 403 Debugging information follows. Path: /drupal8/entity-embed/dialog/entity-embed/full_html StatusText: Forbidden ResponseText: A fatal error occurred: Drupal.AjaxError.prototype = new Error();

src/EntityEmbedController.php
```diff
@@ -0,0 +1,75 @@
...
+use Drupal\filter\FilterFormatInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * Returns responses for EntityEmbed module routes.
+ */
+class EntityEmbedController extends ControllerBase {
+  use EntityHelperTrait;
+
+  /**
```
davereid: I don't think we need any of these request methods or constructors. We just need to add a Request $request parameter to the preview method and that's it.

js/plugins/entityembed/plugin.js
```diff
@@ -33,6 +36,44 @@
...
+        button: 'Insert entity',
+
+        // Generate the preview of the element and render it.
+        upcast: function (element) {
+          var attributes = element.attributes;
+          if (attributes['data-entity-type'] === undefined || (attributes['data-entity-id'] === undefined && attributes['data-entity-uuid'] === undefined)) {
+            return;
+          }
+          var request = {};
+          request['filter-format'] = editor.config.drupal.format;
```
davereid: I think we can remove this query string parameter now that it's in the URL.

cs-shadow: Removed.

cs-shadow:
@slashrsm, the access denied issue is resolved now. Reason was: https://www.drupal.org/node/1906810

davereid:
I'm unable to see anything different with this branch, nor any additional AJAX request when I embed content.

src/EntityEmbedController.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\entity_embed\EntityEmbedController
+ */
+
+namespace Drupal\entity_embed;
+
+use Drupal\Component\Serialization\Json;
```
davereid: Drupal\Component\Serialization\Json and Symfony\Component\DependencyInjection\ContainerInterface are no longer needed here.

src/EntityEmbedController.php
```diff
@@ -0,0 +1,51 @@
...
+class EntityEmbedController extends ControllerBase {
+
+  /**
+   * Returns an Ajax response to generate preview of an entity.
+   *
+   * Expects the the HTML element as GET parameter.
+   *
+   * @param \Symfony\Component\HttpFoundation\Request $request
+   *   The request object.
+   *
```
davereid: Extra unwanted line break in docblock.

src/EntityEmbedController.php
cs-shadow: Removed.

cs-shadow:
@davereid, ensure that the new preview route is active. (Maybe, try running update.php since this branch introduces a new route)

davereid:
@cs-shadow My preview route is definitely active. I can hit http://drupal8.dev/entity-embed/preview/full_html?value=Test and it returns `{"content":"Test"}`

davereid:
Looks like I was a victim of browser caching. It appears to be working now.

js/plugins/entityembed/plugin.js
```diff
@@ -33,6 +36,43 @@
         }
       });
 
+      // Register the entity embed widget.
+      editor.widgets.add('entityembed', {
+        // Minimum HTML which is required by this widget to work.
+        requiredContent: 'div[data-entity-type,data-entity-uuid]',
+
+        // Add entity embed button
+        button: 'Insert entity',
```
davereid: Does this need Translation?

js/plugins/entityembed/plugin.js
```diff
@@ -33,6 +36,43 @@
         }
       });
 
+      // Register the entity embed widget.
+      editor.widgets.add('entityembed', {
```
davereid: Should be 'entity_embed' (not smooshed string).

slashrsm:
Tested it and it works now.

js/plugins/entityembed/plugin.js
```diff
@@ -33,6 +36,43 @@
...
+        // Minimum HTML which is required by this widget to work.
+        requiredContent: 'div[data-entity-type,data-entity-uuid]',
```
davereid: Looks like this would fail if just data-entity-id was used? Is it possible to relax this restriction?

js/plugins/entityembed/plugin.js
slashrsm: Agree, that would look nicer.

js/plugins/entityembed/plugin.js
slashrsm, line 61: OK. Please open a follow-up so we don't forget about it.

js/plugins/entityembed/plugin.js
cs-shadow: Done.

js/plugins/entityembed/plugin.js
cs-shadow, line 61: Here's the follow-up issue: https://www.drupal.org/node/2291193

js/plugins/entityembed/plugin.js
cs-shadow: Tested, this does not fails when entity-id is used. Not sure on how requiredContent is used because seems like that attribute is not respected by the widget.
EDIT: Seems like i misinterpreted few things. Its used when entity is being embedded. As per discussion with @davereid, making just the `data-entity-type` as required attribute.

js/plugins/entityembed/plugin.js
cs-shadow: On second thoughts, I don't think this line is even required. Same with the icon attribute. Everything works just fine if i remove those nor are they used anywhere IMO.

js/plugins/entityembed/plugin.js
davereid: Doesn't seem to actually be fixed. I see other instances of 'entityembed' like above in the CKEDITOR.plugins.add line.

cs-shadow: Core uses smooshed strings. For instance, `drupalimage` and `drupallink`. So, I'm not sure what's the best way around here. @davereid do you think we should go with `entity_embed` for plugin name anyway?

PR #54, opened 2014-06-19
Add module to Filter package

cs-shadow:


entity_embed.info.yml
```diff
@@ -2,5 +2,6 @@ name: Entity Embed
 type: module
 description: 'Allows entities to be rendered using a text format.'
 core: 8.x
+package: Filter
```
davereid: Should probably be 'Filters' since core seems to use a plural for categories that contrib could also use, like 'Field types'

PR #55, opened 2014-06-21
Simpletests to ensure filter is working properly.

cs-shadow:
d.o issue: https://www.drupal.org/node/2290277

### DIFF
--- a/src/Tests/EntityEmbedFilterTest.php
+++ b/src/Tests/EntityEmbedFilterTest.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * @file
+ * contains \Drupal\entity_embed\Tests\EntityEmbedFilterTest.
+ */
+
+namespace Drupal\entity_embed\Tests;
+
+use Drupal\simpletest\WebTestBase;
+
+/**
+ * Tests the entity_embed filter.
+ */
+class EntityEmbedFilterTest extends WebTestBase {
+
+  /**
+   * Modules to enable.
+   *
+   * @var array
+   */
+  public static $modules = array('entity_embed', 'filter', 'node');
+
+  public static function getInfo() {
+    return array(
+      'name' => 'Entity Embed Filter Test',
+      'description' => 'Tests the entity_embed filter',
+      'group' => 'Entity Embed',
+    );
+  }
+
+  protected function setUp() {
+    parent::setUp();
+
+    // Create a page content type.
+    $this->drupalCreateContentType(array('type' => 'page', 'name' => 'Basic page'));
+
+    // Create Filtered HTML text format and enable entity_embed filter.
+    $filtered_html_format = entity_create('filter_format', array(
+      'format' => 'filtered_html',
+      'name' => 'Filtered HTML',
+      'filters' => array(
+        'entity_embed' => array(
+          'status' => 1,
+        ),
+      ),
+    ));
+    $filtered_html_format->save();
+
+    // Create a user with required permissions.
+    $this->webUser = $this->drupalCreateUser(array(
+      'access content',
+      'create page content',
+      'use text format filtered_html',
+    ));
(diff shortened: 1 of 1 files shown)


jibran:
I think this can only be tested by https://www.drupal.org/node/2288871#comment-8895215

src/Tests/EntityEmbedFilterTest.php
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  protected function setUp() {
+    parent::setUp();
+
+    // Create a page content type.
+    $this->drupalCreateContentType(array('type' => 'page', 'name' => 'Basic page'));
+
+    // Create Filtered HTML text format and enable entity_embed filter.
+    $filtered_html_format = entity_create('filter_format', array(
```
jibran: This should be full html. Drupal 8 don't support divs in filtered html.

src/Tests/EntityEmbedFilterTest.php
```diff
@@ -0,0 +1,154 @@
...
+    $settings = array();
+    $settings['type'] = 'page';
+    $settings['title'] = 'Test entity embed with entity-uuid and view-mode';
+    $settings['body'] = array(array('value' => $content));
+    $node = $this->drupalCreateNode($settings);
+
+    $html = $this->drupalGet('node/' . $node->id());
+
+    $this->assertText($this->embedContent, 'Embedded node exists in page.');
+    $this->assertNoText('This placeholder should not be rendered.', 'Placeholder does not appears in the output when embed is successful.');
```
jibran: how about using `$content` instead of 'This placeholder should not be rendered.'

src/Tests/EntityEmbedFilterTest.php
cs-shadow: Umm, this $filtered_html_format is not exactly Drupal's restricted html format. Its a format that we are defining. And since we have not enabled the `filter_html` filter which restricts html tags, this format is essentially full html with the entity_embed filter enables.
I can verify that this works since I've tested in on my local installation.

src/Tests/EntityEmbedFilterTest.php
cs-shadow: Nice idea, I'll replace all such occurrences.

cs-shadow:
@jibran, unless we've the travis-ci, we can test this on our local installations using the `run-tests.sh` script.

src/Tests/EntityEmbedFilterTest.php
slashrsm: Maybe rename this format to prevent confusion.

cs-shadow: OK, what about `$full_html` since essentially its full html only?

slashrsm: I'd go with something that is completely unrelated to filters that ship with core.

src/Tests/EntityEmbedFilterTest.php
cs-shadow: (Somehow my earlier comment got removed, so posting again).
This might not be a good idea, since assertText excepts plain text as argument and `$content` is a HTML string, so I think we should not use `$content` here.

src/Tests/EntityEmbedFilterTest.php
cs-shadow: OK, naming stuff is always difficult for me.

cs-shadow: Renamed to `custom_html` format.

src/Tests/EntityEmbedFilterTest.php
slashrsm: what about {code}strip_tags($content){code}?

cs-shadow: Done.

PR #56, opened 2014-06-21
Removed modification of data attribute keys.

davereid:
Was starting to get frustrated about sometimes we have to use 'data-entity-type' and other times we use 'entity-type'. It would be better for us and for others to just be consistent and not remove the 'data-' prefix from the attributes.


davereid:
Tested and confirmed this is working as expected.

PR #58, opened 2014-06-29
Adds travis support.

nickschuch:


.travis.yml
```diff
@@ -0,0 +1,66 @@
...
+  - sed -i '1i export PATH="$HOME/.composer/vendor/bin:$PATH"' $HOME/.bashrc
+  - source $HOME/.bashrc
+  - composer self-update
+
+  # Drush.
+  - composer global require "youngj/httpserver:dev-master#41dd2b7"
+  - composer global require drush/drush:dev-master
+
+  # Codesniffer.
+  - pyrus install pear/PHP_CodeSniffer
```
jibran: Why are we not installing it using composer?

.travis.yml
```diff
@@ -0,0 +1,66 @@
...
+  - sudo apt-get install -y --force-yes php5-cgi php5-mysql
+
+  # Move the checked out module into the correct structure.
+  - mkdir /tmp/$MODULE_NAME
+  - mv * /tmp/$MODULE_NAME/
+  - git clone --branch $DRUPAL_VERSION $DRUPAL_REPO drupal
+  - mv /tmp/$MODULE_NAME drupal/modules/
+
+  # Clone Coder module for later.
+  - git clone --branch $CODER_VERSION $CODER_REPO drupal/modules/coder
```
jibran: Why are we not cloning it in ~/.drush dir?

.travis.yml
```diff
@@ -0,0 +1,66 @@
...
+php:
+  - 5.4
+  - 5.5
+
+env:
+  global:
+    - MODULE_NAME='entity_embed'
+    - MODULE_TEST_GROUP='Entity Embed'
+    - DRUPAL_REPO='git://drupalcode.org/project/drupal.git'
+    - DRUPAL_VERSION='8.0-alpha12'
```
cs-shadow: This module will NOT work with the last alpha release. There are couple of issues in core which are only in the latest dev release on which this module depends. So, we'll need the latest dev release until there's a new alpha release.

.travis.yml
nickschuch: Good idea.

.travis.yml
nickschuch: It could, given it is a full module I just put it in here.

.travis.yml
nickschuch: Awesome, we can specify the branch instead in this case. Ideally I think we would want to bump  releases when they come out so we are chasing releases and not commit's.

cs-shadow: Yes, that's exactly what we want, but there were a couple of commits that were crucial for development of this module and hence this dependency.

davereid: I would prefer if we know that the module no longer works with D8 HEAD so we could fix issues.

.travis.yml
davereid: Please take a look at https://github.com/fago/rules/blob/8.x-3.x/.travis.yml which just symlinks the module directory. That seems better.

nickschuch: Why does it seem better? I think they both acheive the same thing. So Im not sure I follow.

.travis.yml
nickschuch: We can switch it to HEAD. That's not a big deal since we have these variables right at the top of the file and don't need to go digging through code at the bottom.

.travis.yml
nickschuch: Am I right in saying there is recursion in that file since the base is symlinked into the Drupal subdir?

jibran: Indeed.

.travis.yml
```diff
@@ -0,0 +1,69 @@
+language: php
+
+php:
+  - 5.4
+  - 5.5
+
+env:
+  global:
+    - MODULE_NAME='entity_embed'
+    - MODULE_TEST_GROUP='Entity Embed'
```
jibran: entity_embed now not Entity Embed

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+  # Ensure we have the latest sources.
+  - sudo apt-get -y update
+
+  # Composer.
+  - sed -i '1i export PATH="$HOME/.composer/vendor/bin:$PATH"' $HOME/.bashrc
+  - source $HOME/.bashrc
+  - composer self-update
+
+  # Drush.
+  - composer global require "youngj/httpserver:dev-master#41dd2b7"
```
jibran, line 30: why not only dev-master?

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+  - git clone --branch $DRUPAL_VERSION $DRUPAL_REPO drupal
+  - mv /tmp/$MODULE_NAME drupal/modules/
+
+before_script:
+  # This fixes a fail when install Drupal.
+  - echo 'sendmail_path = /bin/true' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+
+  # Install Drupal and enable the required modules (including this one).
+  - mysql -e 'create database drupal;'
+  - cd $TRAVIS_BUILD_DIR/drupal && drush --yes site-install minimal --db-url="mysql://root@127.0.0.1/drupal"
```
jibran: I think we could use `testing` profile instead of `minimal`

.travis.yml
nickschuch, line 30: Drush's output recommends this as the preferred version of the library.

.travis.yml
nickschuch: no worries, will update

.travis.yml
nickschuch: Yep, I think your right there, will be much quicker, will update.

.travis.yml
```diff
@@ -0,0 +1,69 @@
...
+  - phpenv rehash
+
+install:
+  # Basic PHP packages.
+  - sudo apt-get install -y --force-yes php5-cgi php5-mysql
+
+  # Move the checked out module into the correct structure.
+  - mkdir /tmp/$MODULE_NAME
+  - mv * /tmp/$MODULE_NAME/
+  - git clone --branch $DRUPAL_VERSION $DRUPAL_REPO drupal
```
jibran, line 50: Suggested by @Berdir You can use --depth 1 to speed up the git clone and only get the current version and not all of the history too.

cs-shadow:
@nickschuch @jibran What else is required here? Any blockers?

Looks good me (and I'm also in favor of adding the depth argument to git clone).

PR #59, opened 2014-06-30
Add tests to ensure preview route/controller is working correctly

cs-shadow:


src/Tests/EntityEmbedPreviewTest.php
```diff
@@ -0,0 +1,105 @@
...
+    ));
+    $format->save();
+
+    // Create a user with required permissions.
+    $this->webUser = $this->drupalCreateUser(array(
+      'access content',
+      'create page content',
+      'use text format custom_format',
+    ));
+    $this->drupalLogin($this->webUser);
```
jibran, line 58: Why not create a EntityEmbedTestBase and move all setup logic(custom format and user creation) there?

cs-shadow, line 58: Nice Idea! Will do that once this PR goes in to maintain coherency.

cs-shadow, line 58: Follow-up issue to move common setup logic to a base class: https://www.drupal.org/node/2295979

PR #60, opened 2014-07-01
Move common setUp functions to a base class

cs-shadow:
d.o issue: https://www.drupal.org/node/2295979

### DIFF
--- a/src/Tests/EntityEmbedTestBase.php
+++ b/src/Tests/EntityEmbedTestBase.php
@@ -1,0 +1,58 @@
+<?php
+
+/**
+ * @file
+ * contains \Drupal\entity_embed\Tests\EntityEmbedTestBase.
+ */
+
+namespace Drupal\entity_embed\Tests;
+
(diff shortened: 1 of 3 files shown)


src/Tests/EntityEmbedTestBase.php
```diff
@@ -0,0 +1,78 @@
...
+   * @var array
+   */
+  public static $modules = array('editor', 'entity_embed', 'filter', 'node');
+
+  /**
+   * Sample content to be used as body of the node to be embedded.
+   *
+   * @var string
+   */
+  protected $embedContent = 'This node is to be used for embedding in other nodes.';
```
davereid: I don't think we need to store this in the class. We could retrieve it using $this->node->body->value() or something right?

jibran:
:+1: for this.

src/Tests/EntityEmbedTestBase.php
cs-shadow: Done.

PR #62, opened 2014-07-03
Use save callback to create widget

cs-shadow:
Details at d.o issue: https://www.drupal.org/node/2297507


davereid:
Yay! I thought it was really odd that we had these additional classes when the core link/image CKEditor plugins didn't need theme.

:clap:

PR #64, opened 2014-07-08
Add 'edit' option for embedded entity's CKEditor widget

cs-shadow:
d.o issue: https://www.drupal.org/node/2295943


slashrsm:
I positioned embeded entity between two paragraphs. When I edited it the one after embeded entity was removed. 

Also... would be great to have some kind of "edit" icon when hovering and entity. Something similar to what core has for embeded images.

cs-shadow:
@slashrsm the issue where the following paragraph was removed is fixed now.
Also, now the edit dialog pops up when you double-click on the widget, which is done by all core widgets.
Though, I'm not sure what you meant by the edit icon? I can't see any icon on embedded images.

PR #71, opened 2014-07-18
Add text align attribute

cs-shadow:
d.o issue: https://www.drupal.org/node/2304425

### DIFF
--- a/src/Form/EntityEmbedDialog.php
+++ b/src/Form/EntityEmbedDialog.php
@@ -83,6 +83,7 @@ class EntityEmbedDialog extends FormBase {
       'data-entity-id' => '',
       'data-entity-embed-display' => 'default',
       'data-entity-embed-settings' => array(),
+      'data-text-align' => 'none',
     );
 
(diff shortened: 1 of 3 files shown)


src/Form/EntityEmbedDialog.php
```diff
@@ -180,7 +181,19 @@ public function buildForm(array $form, array &$form_state, FilterFormat $filter_
           $display->setContextValue('entity', $entity);
           $form['attributes']['data-entity-embed-settings'] += $display->buildConfigurationForm($form, $form_state);
         }
-        // @todo Re-add caption and alignment attributes.
+        $form['attributes']['data-text-align'] = array(
+          '#title' => $this->t('Align'),
```
slashrsm, line 185: #description would be great here.

cs-shadow, line 185: This form element is copied exactly from the DrupalImageDialog, and that does not has a #description. Hence, I feel its better to keep it this way for a unified UX.

PR #72, opened 2014-07-18
Use uppercase TRUE instead of small case

cs-shadow:
Minor coding standard nitpick.


cs-shadow:
Cool, The Travis CI build passed without complaining for anything.

PR #73, opened 2014-07-18
Use class names as typehint in function definitions instead of its fully qualified name

cs-shadow:
d.o issue: https://www.drupal.org/node/2305895


jibran:
Perhaps you can add a use statement above. Just a suggestion.

davereid:
These are hooks, and this matches the current Drupal core documentation standards. I don't believe this should be changed.

cs-shadow:
Ah, my bad. Didn't knew that about hooks.

Thanks for pointing that out.

PR #75, opened 2014-07-19
Add tests for entity reference field formatter display

cs-shadow:
d.o issue: https://www.drupal.org/node/2305897


src/Tests/EntityEmbedDisplayPluginTest.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+/**
+ * @file
+ * contains \Drupal\entity_embed\Tests\EntityEmbedDisplayPluginTest.
```
jibran: Contains

cs-shadow: Nice Catch. Noticed that this is also used in other files in the Tests directory, fixed those also.

PR #78, opened 2014-07-20
Add tests for image field formatter plugin

cs-shadow:
d.o issue: https://www.drupal.org/node/2305903

### DIFF
--- a/src/Tests/EntityEmbedImageFieldFormatterTest.php
+++ b/src/Tests/EntityEmbedImageFieldFormatterTest.php
@@ -46,9 +46,23 @@ class EntityEmbedImageFieldFormatterTest extends EntityEmbedTestBase {
    */
   public function testDisplayPluginOptions() {
     $plugin_options = $this->displayPluginManager()->getDefinitionOptionsForEntity($this->image);
-    print_r($plugin_options);
     // Test that 'image:image' plugin is available.
     $this->assertTrue(array_key_exists('image:image', $plugin_options), "The 'Image' plugin is available.");
+  }
+
+  /**
+   * Tests that image plugin is not available to files other than images.
+   */
+  public function testDisplayPluginOptionsWithFile() {
+    file_put_contents('public://example.txt', $this->randomName());
+    $file = entity_create('file', array(
+      'uri' => 'public://example.txt',
+    ));
+    $file->save();
+
+    $plugin_options = $this->displayPluginManager()->getDefinitionOptionsForEntity($file);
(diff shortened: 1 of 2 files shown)


src/Tests/EntityEmbedImageFieldFormatterTest.php
```diff
@@ -0,0 +1,109 @@
...
+   * Created file entity.
+   *
+   * @var \Drupal\file\Entity\File
+   */
+  protected $image;
+
+  protected function setUp() {
+    parent::setUp();
+
+    file_unmanaged_copy(DRUPAL_ROOT . '/core/misc/druplicon.png', 'public://example.png');
```
davereid: We should be able to use drupalGetTestFiles('image') since it handles some of this for us. See FileMoveTest::testNormal():

``` php
$file = entity_create('file', (array) current($this->drupalGetTestFiles('image')));
```

src/Tests/EntityEmbedImageFieldFormatterTest.php
```diff
@@ -0,0 +1,109 @@
...
+    $plugin_options = $this->displayPluginManager()->getDefinitionOptionsForEntity($this->image);
+    // Test that 'image:image' plugin is available.
+    $this->assertTrue(array_key_exists('image:image', $plugin_options), "The 'Image' plugin is available.");
+  }
+
+  /**
+   * Tests that image plugin is not available to files other than images.
+   */
+  public function testDisplayPluginOptionsWithFile() {
+    file_put_contents('public://example.txt', $this->randomName());
```
davereid: See same comment about using drupalGetTestFiles()

davereid:
I feel like we're sneaking EntityEmbed in the name of the test classes a bit. I think this should be named ImageFieldFormatterTest.

davereid:
I also feel like we could condense all the tests into one method, rather than all the separate test methods which require the D8 environment to be reinstalled. I don't really see anything that needs that environment separation.

src/Tests/EntityEmbedImageFieldFormatterTest.php
cs-shadow: Well, if we use `drupalGetTestFiles()`, we'll have to explicitly move the file to `Public://` for embeds to work. So, I don't fell it'll do much good. Also, there are a lot of instances in core where files are created this (the original) way.

cs-shadow:
Moved all tests inside a single method and renamed the test. The latter came with a sigh of relief as the names were getting ridiculously large.

PR #81, opened 2014-07-21
Add tests for entity_embed hooks

cs-shadow:
d.o issue: https://www.drupal.org/node/2306645


cs-shadow:
Hrm, test failures, still it's green :/

cs-shadow:
Really strange that the tests pass locally and on d.o: https://www.drupal.org/node/2306645

Maybe some glitch with travis.

PR #82, opened 2014-07-22
Use custom tag for container DOM element

cs-shadow:
d.o issue: https://www.drupal.org/node/2307553


cs-shadow:
Merged this since without this, the CKEditor integration was broken in HEAD. Feel free to open any follow-ups if required.

PR #86, opened 2014-08-05
Rename tag of container elemet to div if it was drupal-entity

cs-shadow:
d.o issue: https://www.drupal.org/node/2307553
Follow-up to: #82

Since it's not possible to rename a DOM node directly, first a new node is created and all attributes/children of parent node are copied to the new node. Then we replace the original node with the new node.


davereid:
Might be better to put this code into setDomNodeContent()

PR #87, opened 2014-08-06
Updated for https://www.drupal.org/node/2310411.

davereid:


davereid:
Somehow the embed form is skipping the second 'embed options' step of the form. Trying to figure out why.

cs-shadow:
@davereid Same thing was happening with me when I was trying this.

src/Form/EntityEmbedDialog.php
```diff
@@ -316,4 +282,68 @@ public function goBack(array &$form, array &$form_state) {
...
 
+  /**
+   * Form submission handler that selects an entity and display embed settings.
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param FormStateInterface $form_state
+   *   An associative array containing the current state of the form.
+   */
+  public function selectEntity(array &$form, FormStateInterface $form_state) {
```
slashrsm: It appears this two can't be in main submit. It is also nicer if we separate diff buttons into diff functions. 

Not sure what to do with errors handling, though (redundant code in both funcitons). submitForm() is not even called if validation fails so it can't be in there. Not sure if there is any suggested pattern how to handle this in AJAX. Would be nice to look into this but I'm unable ATM.

Selecting, settings, embeding now works again.

davereid:
I'm just going to merge this for now. Obviously something more is wrong with Travis.CI and let's fix that separately rather than hold this up.

PR #88, opened 2014-08-09
Entity embed config entity

cs-shadow:
d.o issue: https://www.drupal.org/node/2273361

@todo:
- Support for customizing button icons.
- ~~Remove entity type field from embed form.~~ [Done]
- Add CRUD tests.

### DIFF
--- a/entity_embed.routing.yml
+++ b/entity_embed.routing.yml
@@ -15,7 +15,7 @@ entity_embed.preview:
     _entity_access: 'filter_format.use'
 
 embed_button.list:
-  path: '/admin/config/config/embed-button'
+  path: '/admin/config/content/embed-button'
   defaults:
     _entity_list: 'embed_button'
     _title: 'Embed Buttons'
--- a/entity_embed.routing.yml
+++ b/entity_embed.routing.yml
@@ -1,9 +1,5 @@
 entity_embed.dialog:
   path: '/entity-embed/dialog/entity-embed/{filter_format}/{embed_button}'
-  options:
-    parameters:
-      embed_button:
-        type: entity:embed_button
   defaults:
     _form: '\Drupal\entity_embed\Form\EntityEmbedDialog'
     _title: 'Embed entity'
--- a/entity_embed.routing.yml
+++ b/entity_embed.routing.yml
@@ -13,3 +13,35 @@ entity_embed.preview:
     _title: 'Preview embedded entity'
   requirements:
     _entity_access: 'filter_format.use'
+
+embed_button.list:
+  path: '/admin/config/config/embed-button'
+  defaults:
+    _entity_list: 'embed_button'
+    _title: 'Embed Buttons'
+  requirements:
+    _permission: 'administer entity embed settings'
+
+embed_button.add:
+  path: '/admin/config/content/embed-button/add'
+  defaults:
+    _entity_form: 'embed_button.add'
+    _title: 'Add Embed Button'
+  requirements:
+    _permission: 'administer entity embed settings'
+
+embed_button.edit:
+  path: '/admin/config/content/embed-button/{embed_button}'
+  defaults:
+    _entity_form: 'embed_button.edit'
+    _title: 'Edit Embed button'
+  requirements:
+    _permission: 'administer entity embed settings'
+
+embed_button.delete:
+  path: '/admin/config/content/embed-button/{embed_button}/delete'
+  defaults:
+    _entity_form: 'embed_button.delete'
+    _title: 'Delete Embed button'
+  requirements:
+    _permission: 'administer entity embed settings'
--- a/entity_embed.routing.yml
+++ b/entity_embed.routing.yml
@@ -24,7 +24,7 @@ embed_button.list:
     _entity_list: 'embed_button'
     _title: 'Embed Buttons'
   requirements:
-    _permission: 'administer entity embed settings'
+    _permission: 'administer embed buttons'
 
 embed_button.add:
   path: '/admin/config/content/embed-button/add'
@@ -32,7 +32,7 @@ embed_button.add:
     _entity_form: 'embed_button.add'
     _title: 'Add Embed Button'
   requirements:
-    _permission: 'administer entity embed settings'
+    _permission: 'administer embed buttons'
 
 embed_button.edit:
   path: '/admin/config/content/embed-button/{embed_button}'
@@ -40,7 +40,7 @@ embed_button.edit:
     _entity_form: 'embed_button.edit'
     _title: 'Edit Embed button'
   requirements:
-    _permission: 'administer entity embed settings'
+    _permission: 'administer embed buttons'
 
 embed_button.delete:
   path: '/admin/config/content/embed-button/{embed_button}/delete'
@@ -48,4 +48,4 @@ embed_button.delete:
     _entity_form: 'embed_button.delete'
     _title: 'Delete Embed button'
   requirements:
-    _permission: 'administer entity embed settings'
+    _permission: 'administer embed buttons'
--- a/entity_embed.routing.yml
+++ b/entity_embed.routing.yml
@@ -1,5 +1,9 @@
 entity_embed.dialog:
-  path: '/entity-embed/dialog/entity-embed/{filter_format}'
+  path: '/entity-embed/dialog/entity-embed/{filter_format}/{embed_button}'
+  options:
+    parameters:
+      embed_button:
+        type: entity:embed_button
   defaults:
     _form: '\Drupal\entity_embed\Form\EntityEmbedDialog'
     _title: 'Embed entity'
--- a/js/plugins/drupalentity/plugin.js
+++ b/js/plugins/drupalentity/plugin.js
@@ -23,13 +23,6 @@
 
           var existingElement = getSelectedEntity(editor);
 
-          var dialogSettings = {
-            title: existingElement ? editor.config.DrupalEntity_dialogTitleEdit : editor.config.DrupalEntity_dialogTitleAdd,
-            dialogClass: 'entity-select-dialog',
-            resizable: false,
-            minWidth: 800
-          };
-
           var existingValues = {};
           if (existingElement && existingElement.$ && existingElement.$.firstChild) {
             var entityDOMElement = existingElement.$.firstChild;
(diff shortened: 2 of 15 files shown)


davereid:
Might make sense just to make the entity label itself the button label? Not sure we need two different properties for labels.

cs-shadow:
@davereid I thought of  a scenario where two different modules are trying to provide for the same entity type and hence may want the same button label, say 'Image' but need to have different entity labels (like 'ModuleX:Image' and 'ModuleY:Image').

I realize that this may not be used very frequently, but I thought it may be good have it customizable. We can do away with it if you think that's better.

src/Plugin/CKEditorPlugin/DrupalEntity.php
```diff
@@ -53,9 +76,23 @@ public function getLibraries(Editor $editor) {
...
+    $buttons = array();
+
+    foreach ($this->embed_buttons as $embed_button) {
+      $button = EmbedButton::load($embed_button);
+      $buttons[$button->label()] = array(
+        'id' => $button->id(),
+        'name' => $button->label(),
+        'label' => $button->getButtonLabel(),
+        'entity_type' => $button->getEntityTypeMachineName(),
+        'image' => url(drupal_get_path('module', 'entity_embed') . '/js/plugins/drupalentity/entity.png'),
```
davereid: Let's make this $button->getButtonImage() method so we can eventually convert this to a user input.

davereid:
Shouldn't the entity_embed.dialog route be changed to require a button entity? Then we could automatically derive the entity type from that. We shouldn't be able to hit the form without a button ID.

js/plugins/drupalentity/plugin.js
```diff
@@ -91,13 +98,19 @@
...
       if (editor.ui.addButton) {
-        editor.ui.addButton('DrupalEntity', {
-          label: Drupal.t('Entity'),
-          command: 'editdrupalentity',
-          icon: this.path + '/entity.png',
-        });
+        for (var key in editor.config.DrupalEntity_buttons) {
+          var button = editor.config.DrupalEntity_buttons[key];
+          editor.ui.addButton(button.name, {
+            label: button.label,
```
davereid, line 105: I wonder if we need to check and make sure this value is sanitized.

src/Entity/EmbedButton.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function getEntityTypeMachineName() {
+    return $this->entity_type;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getEntityTypeLabel() {
+    return $this->entity_types[$this->entity_type];
```
davereid: I think it makes more sense to do $this->entityManager()-> getDefinition($this->entity_type)->getLabel(); rather than needing to store all entity type labels.

src/Entity/EmbedButton.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function getButtonLabel() {
+    return $this->button_label;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getButtonImage() {
+    return url(drupal_get_path('module', 'entity_embed') . '/js/plugins/drupalentity/entity.png');
```
davereid: I think this should be using file_create_url() instead of url()? I don't think url() is typically used for paths to actual files.

src/Form/EmbedButtonDeleteForm.php
```diff
@@ -0,0 +1,50 @@
...
+  public function getConfirmText() {
+    return $this->t('Delete');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submit(array $form, FormStateInterface $form_state) {
+    $this->entity->delete();
+    drupal_set_message($this->t('Category %label has been deleted.', array('%label' => $this->entity->label())));
```
davereid: Button, not category.

src/Entity/EmbedButton.php
cs-shadow: OK, I didn't knew that. Done.

src/Entity/EmbedButton.php
cs-shadow: Definitely, it bugged me from the start.

src/Form/EmbedButtonDeleteForm.php
cs-shadow: Done

js/plugins/drupalentity/plugin.js
cs-shadow, line 105: Added a checkPlain in `\Drupal\entity_embed\Plugin\CKEditorPlugin\DrupalEntity` for all fields.

entity_embed.routing.yml
```diff
@@ -1,5 +1,9 @@
 entity_embed.dialog:
-  path: '/entity-embed/dialog/entity-embed/{filter_format}'
+  path: '/entity-embed/dialog/entity-embed/{filter_format}/{embed_button}'
+  options:
+    parameters:
+      embed_button:
```
davereid: I'm a bit curious why embed_button is defined here, but not filter_format?

cs-shadow: `filter_format` was already upcasted and not `embed_button` (though I'm not sure why). Without this definition, it was treated as a string and not as an entity.

src/Form/EntityEmbedDialog.php
```diff
@@ -172,6 +172,14 @@ public function buildForm(array $form, FormStateInterface $form_state, FilterFor
           '#prefix' => '<div id="data-entity-embed-settings-wrapper">',
           '#suffix' => '</div>',
         );
+        $form['attributes']['data-button-entity'] = array(
```
davereid: Nitpick, but let's name this data-embed-button

davereid: Also would this mean we don't need to store data-entity-type in the embed?

cs-shadow: Possibly yes, we should be able to get `data-entity-type` given the id of embed button. But I say let's keep it separate for clarity.

src/Form/EmbedButtonForm.php
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Constructs a new EmbedButtonForm.
+   *
+   * @param \Drupal\Core\Entity\Query\QueryFactory $entity_query
+   *   The entity query.
+   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   *   The entity manager service.
+   */
+  public function __construct(QueryFactory $entity_query, EntityManagerInterface $entity_manager) {
+    $this->entityQuery = $entity_query;
```
jibran, line 43: This is not defined in this class.

src/Plugin/CKEditorPlugin/DrupalEntity.php
```diff
@@ -22,15 +24,37 @@
...
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    $this->embed_buttons = \Drupal::entityQuery('embed_button')->execute();
```
jibran, line 47: I think we should declare this properties.

src/Plugin/CKEditorPlugin/DrupalEntity.php
```diff
@@ -53,9 +77,23 @@ public function getLibraries(Editor $editor) {
    * {@inheritdoc}
    */
   public function getConfig(Editor $editor) {
+    $buttons = array();
+
+    foreach ($this->embed_buttons as $embed_button) {
+      $button = EmbedButton::load($embed_button);
+      $buttons[$button->label()] = array(
+        'id' => String::checkPlain($button->id()),
```
jibran, line 93: I don't think it check_plain is not required as it is a machine name. But I do agree user can import config data so perhaps this is a good idea.

src/Plugin/CKEditorPlugin/DrupalEntity.php
```diff
@@ -53,9 +77,23 @@ public function getLibraries(Editor $editor) {
...
+        'name' => String::checkPlain($button->label()),
+        'label' => String::checkPlain($button->getButtonLabel()),
+        'entity_type' => String::checkPlain($button->getEntityTypeMachineName()),
```
jibran, line 96: Same as above.

src/Plugin/CKEditorPlugin/DrupalEntity.php
```diff
@@ -53,9 +77,23 @@ public function getLibraries(Editor $editor) {
...
+        'image' => String::checkPlain($button->getButtonImage()),
```
jibran, line 97: same as above.

jibran, line 97: If we do need this then we can move this getter function.

src/Plugin/CKEditorPlugin/DrupalEntity.php
jibran, line 96: If we do need this then we can move this getter function.

src/Plugin/CKEditorPlugin/DrupalEntity.php
jibran, line 93: @davereid please advise.

entity_embed.routing.yml
jibran: @dawehner could you please help here.

src/Form/EmbedButtonForm.php
cs-shadow, line 43: Done.

src/Plugin/CKEditorPlugin/DrupalEntity.php
cs-shadow, line 47: Done.

src/Plugin/CKEditorPlugin/DrupalEntity.php
cs-shadow, line 93: I think check_plain() should be used, though I'm not sure whether to use it here or in the get functions.

src/Entity/EmbedButton.php
```diff
@@ -0,0 +1,108 @@
...
+   */
+  public $entity_type;
+
+  /**
+   * Overrides ConfigEntityBase::__construct().
+   */
+  public function __construct(array $values, $entity_type) {
+    parent::__construct($values, $entity_type);
+
+    $this->entity_types = $this->entityManager()->getEntityTypeLabels();
```
jibran: Any reason in add/edit form these are grouped and note here. If it doesn't matter then we should grouped these here as well. i.e. `getEntityTypeLabels(TRUE)`

jibran:
@cs-shadow overall amazing work.

src/Entity/EmbedButton.php
cs-shadow: Well, that was obsolete code so I removed it. Thanks for pointing it out.

cs-shadow:
@jibran thanks!

entity_embed.routing.yml
slashrsm: Any success with this?

cs-shadow: OK, automatic upcasting is working fine now (like it should have earlier also). I think there was something else which I did wrong before. But in any case, this is working now.

PR #91, opened 2014-08-27
Make icon for embed buttons customizable

cs-shadow:
d.o issue: https://www.drupal.org/node/2324197

### DIFF
--- a/config/schema/embed_button.schema.yml
+++ b/config/schema/embed_button.schema.yml
@@ -15,5 +15,5 @@ entity_embed.embed_button.*:
       type: string
       label: 'Entity type'
     icon_fid:
-      type: string
+      type: integer
       label: 'Button image'
--- a/config/schema/embed_button.schema.yml
+++ b/config/schema/embed_button.schema.yml
@@ -1,6 +1,14 @@
+entity_embed.settings:
+  type: mapping
+  label: 'Entity embed settings'
+  mapping:
+    upload_directory:
+      type: string
+      label: 'Upload directory for buton images'
+
 entity_embed.embed_button.*:
   type: config_entity
-  label: 'Entity button config'
+  label: 'Embed button config'
   mapping:
     label:
       type: label
--- a/config/schema/embed_button.schema.yml
+++ b/config/schema/embed_button.schema.yml
@@ -22,6 +22,6 @@ entity_embed.embed_button.*:
     entity_type:
       type: string
       label: 'Entity type'
-    icon_fid:
+    button_icon_fid:
       type: integer
       label: 'Button image'
--- a/config/schema/embed_button.schema.yml
+++ b/config/schema/embed_button.schema.yml
@@ -14,3 +14,6 @@ entity_embed.embed_button.*:
     entity_type:
       type: string
       label: 'Entity type'
+    icon_fid:
+      type: string
+      label: 'Button image'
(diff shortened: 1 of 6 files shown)


config/schema/embed_button.schema.yml
```diff
@@ -14,3 +14,6 @@ entity_embed.embed_button.*:
     entity_type:
       type: string
       label: 'Entity type'
+    icon_fid:
+      type: string
```
slashrsm: integer

src/Form/EmbedButtonForm.php
```diff
@@ -92,6 +95,18 @@ public function form(array $form, FormStateInterface $form_state) {
       '#description' => $this->t("Label for the button to be shown in CKEditor toolbar."),
       '#required' => TRUE,
     );
+    $form['icon_fid'] = array(
+      '#title' => $this->t('Button image'),
+      '#type' => 'managed_file',
+      '#description' => $this->t("Image for the button to be shown in CKEditor toolbar. Leave empty to use the default Entity icon."),
+      '#upload_location' => 'public://embed_buttons/',
```
slashrsm: make this configurable. No need to create UI, but still get it from config system.

davereid: Do we need to check if this saves the uploaded file as permanent or does it stay a temporary file.?

cs-shadow: It's permanent AFAICT.

cs-shadow: @slashrsm Done.

config/schema/embed_button.schema.yml
```diff
@@ -14,3 +22,6 @@ entity_embed.embed_button.*:
     entity_type:
       type: string
       label: 'Entity type'
+    icon_fid:
```
davereid: Bikeshedding; In keeping with the naming of button_label, I think button_icon_fid would make more sense.

cs-shadow: Done. Anything else that doesn't seems right?

src/Form/EmbedButtonForm.php
```diff
@@ -59,7 +59,12 @@ public static function create(ContainerInterface $container) {
    */
   public function form(array $form, FormStateInterface $form_state) {
     $form = parent::form($form, $form_state);
+
     $embed_button = $this->entity;
+    $button_image = $embed_button->button_icon_fid ? array($embed_button->button_icon_fid) : array();
+    $upload_directory = \Drupal::config('entity_embed.settings')->get('upload_directory');
```
slashrsm, line 66: I'd make entire location configurable (i.e. you get public://foo/bar instead of foo/bar). This automatically makes file scheme configurable also. 

Besides that looks OK. If you fix that I'm fine to merge.

src/Form/EmbedButtonForm.php
```diff
@@ -59,7 +59,13 @@ public static function create(ContainerInterface $container) {
    */
   public function form(array $form, FormStateInterface $form_state) {
     $form = parent::form($form, $form_state);
+
     $embed_button = $this->entity;
+    $button_image = $embed_button->button_icon_fid ? array($embed_button->button_icon_fid) : array();
+    $file_scheme = \Drupal::config('entity_embed.settings')->get('file_scheme');
```
slashrsm, line 65: Is there any real benefit for separating those two? I'd go with something like:

```
\Drupal::config('entity_embed.settings')->get('upload_location');
```

cs-shadow, line 65: It's just the way this is done in most places in Core. Also, given the fact that the scheme and location are independent of each other, it'll be easier to swap one thing with the other if we want to do so. IMO the code looks more modular this way though there's no benefit per se. Let me know if this version works.

PR #93, opened 2014-09-02
add composer.json

dawehner:
https://www.drupal.org/node/2331489


composer.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "drupal/entity_embed",
+  "description": "Allows entities to be rendered using a text format.",
+  "type": "drupal-module",
```
davereid, line 4: If we're using the custom 'type' information, why aren't we also requiring composer/installers as per https://getcomposer.org/doc/faqs/how-do-i-install-a-package-to-a-custom-path-for-my-framework.md ?

composer.json
```diff
@@ -0,0 +1,9 @@
...
+  "license": "GPL-2.0+",
+  "minimum-stability": "dev",
```
davereid, line 6: Don't think we need this line?

composer.json
```diff
@@ -0,0 +1,9 @@
...
+  "require": { }
```
davereid, line 7: Do we need to have an empty require list, or can this just be excluded?

composer.json
dawehner, line 4: oh mh, I haven't thought of requiring it. we do actually have that composer already part of our main composer.json file, see https://blog.erdfisch.de/node/412

composer.json
dawehner, line 6: mh, github doesn't tell me which one you exactly meant. the license or the minimum-stablity?

composer.json
dawehner, line 7: Technically you should at least add the required php version.

composer.json
davereid, line 6: minimum-stability - seems only required if we had requirements, but we don't (and won't) have any

composer.json
davereid, line 7: We don't need any additional requirement other than the version required by Drupal core, so I don't understand.

ericduran:
I incorporated this PR with -> https://github.com/drupal-media/entity_embed/pull/94

I brought in this commit as is (https://github.com/ericduran/entity_embed/commit/af1418905672fb022d4c4ac60d75ef45ea2f4291) and made my changes on top of it. To make sure the history and committer is correct. :+1:

cs-shadow:
Closing this PR in favor of #94.

davereid:
I still had open questions about certain things about this composer.json file though.

cs-shadow:
@davereid Maybe we can move the discussion to the new PR. Feel free to reopen this PR if you like that better.

ericduran:
@davereid I address all your concern on #94.

PR #94, opened 2014-09-06
Fixing travis & Properly fixing composer for travis-ci

ericduran:
I did a couple of things in this PR but all with the end goal to fix travis-ci. I also cleaned some stuff up.
- I needed a composer.json file so I git am (#93) for proper git commit credits (cc @dawehner)
- Added proper dev requirements for the module (drush, phpcs, coder)
- Kill before_install , this shouldn't be needed anymore 
- Switch drush, phpcs, coder to use the vendor/bin binaries
- Change phpcs to ignore the vendor dir, and make sure it test against the proper filetypes
- Switch to using drush test-run over run-test.sh since it reports proper exit codes
- last but not least, I run the drush run-server with nohup so we don't have to worry about the output eliminating the need for netstat.

I tested most of this locally but it lets see what travis thinks. This should also improve the travis build time a bit.


ericduran:
Ha, I forgot "test-run".

cs-shadow:
phpcs produces some false positives and we've discussed before and hence we do not want to fail a build if phpcs fails. So, it'll be better if could run phpcs but not fail due to it. I'm not sure what's the best way here, but maybe we can exit just based on the return code of test-run.

ericduran:
@cs-shadow That makes sense, well I can set it up so phpcs exit code is ignore.

ericduran:
@cs-shadow in my experience they're not really false-positive, it just considers any violation a failure.

ericduran:
Ok this PR is good to go. Travis-CI is reporting failure but that is a correct failure.

Those test are failing locally too. I also set up phpcs to always return true so as not to fail because of coding standard as requested by @cs-shadow.

composer.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "drupal/entity_embed",
+  "description": "Allows entities to be rendered using a text format.",
+  "license": "GPL-2.0+",
+  "require-dev": {
```
dawehner, line 5: Well, technically this also requires a specific PHP version and some version of Drupal (of which the later doesn't work 100% perfect in case you build a drupal site: https://blog.erdfisch.de/node/412)

ericduran, line 5: @dawehner Technically that's not needed and won't really affect anything.

I'm using require-dev and not require to only declare the dependencies needed for development. Since composer is not being used to set up the dependencies needed for this project to work on a website there's no need to declare require and we can leave that to the regular php version in the info file.

dawehner, line 5: Maybe this is true for your usecase though, as described in the article, we
do actually use composer to built our drupal project itself (similar to
drush makefiles ) in which case having a proper require would be helpful

ericduran, line 5: @dawehner I think there's a misunderstanding.

The require property is still not require to work with composer installers. Neither us the php-version. 

We can add the project type because that is require but that's it.

dawehner, line 5: Okay, deal :)

davereid:
I think we're just going to close this now that we've already added a .gitignore, composer.json, and switched to using drupal_ti for the .travis.yml file.

PR #99, opened 2014-10-04
Use route based URL generation

cs-shadow:
https://www.drupal.org/node/2346779 deprecated url() method and it's recommended to use route based Url generation instead.


slashrsm:
Test fails. Do we need to chase HEAD?

cs-shadow:
@slashrsm Yes, all field formatter-related tests are failing.

slashrsm:
OK. Let's merge.

PR #100, opened 2014-10-21
Change EntityEmbedDisplay plugin namespace and move plugins

Berdir:
https://www.drupal.org/node/2360645

### DIFF
--- a/src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
@@ -33,7 +33,7 @@ class EntityEmbedDisplayManager extends DefaultPluginManager {
    *   The module handler to invoke the alter hook with.
    */
   public function __construct(\Traversable $namespaces, CacheBackendInterface $cache_backend, ModuleHandlerInterface $module_handler) {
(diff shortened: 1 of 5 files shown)


jibran:
Wow we have #100 PRs progress++.

slashrsm:
Looks good to me. Test are failing, but it is same in HEAD.

davereid:
Hrm, this was actually done on purpose. Having 'Plugin' in the namespace is a Drupalism and isn't necessary for implementers. It was also nice to keep the code in the same place, rather than in two different places.

Berdir:
Yes, this is a Drupalism. But not every Drupalism is bad. This pattern makes it easy to look at a module at see within seconds with which other modules it is integrating, and what kind of plugins it is implementing. If every module makes up its own patterns, then that is a lot harder to see.

And I think it is also a good thing to have the base class, interface and manager separated from the actual plugins.

davereid:
Shouldn't the namespace be Plugin/entity_embed/Display then? Because having both 'entity_embed' and 'EntityEmbedDisplay' seems like unnecessary duplication.

jibran:
+1 for just display directory structure and leave Annotation as is. Views uses the same apporoch with its plugins but Annotations are @ViewsDisplay, @ViewsStyle and etc.

Berdir:
Fine with me, core has both ways, there is also Field/FieldWidget/Type/Formatter for example. The advantage is that the directory matches the annotation class.

PR #101, opened 2014-10-21
Add autocomplete for entity selection

Berdir:
https://www.drupal.org/node/2360725

### DIFF
--- a/entity_embed.routing.yml
+++ b/entity_embed.routing.yml
@@ -45,3 +45,11 @@ embed_button.delete:
     _title: 'Delete Embed button'
   requirements:
     _permission: 'administer embed buttons'
+
+entity_embed.autocomplete_entity:
+  path: '/entity-embed/autocomplete/{entity_type_id}'
+  defaults:
+    _controller: '\Drupal\entity_embed\Controller\AutocompleteController::autocompleteEntity'
+  requirements:
+    # @todo What to do here?
+    _permission: 'TRUE'
--- a/entity_embed.routing.yml
+++ b/entity_embed.routing.yml
@@ -47,9 +47,8 @@ embed_button.delete:
     _permission: 'administer embed buttons'
 
 entity_embed.autocomplete_entity:
-  path: '/entity-embed/autocomplete/{entity_type_id}'
+  path: '/entity-embed/autocomplete/{filter_format}/{embed_button}'
   defaults:
     _controller: '\Drupal\entity_embed\Controller\AutocompleteController::autocompleteEntity'
   requirements:
-    # @todo What to do here?
-    _permission: 'TRUE'
+    _entity_access: 'filter_format.use'
--- a/src/Controller/AutocompleteController.php
+++ b/src/Controller/AutocompleteController.php
@@ -9,8 +9,11 @@ namespace Drupal\entity_embed\Controller;
 
 use Drupal\Component\Utility\String;
 use Drupal\Core\Controller\ControllerBase;
+use Drupal\Core\Entity\Query\QueryFactory;
 use Drupal\entity_embed\EmbedButtonInterface;
 use Drupal\filter\Entity\FilterFormat;
+use Drupal\filter\FilterFormatInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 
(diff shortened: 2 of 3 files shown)


davereid:
Sounds like an upstream bug to file in core.

entity_embed.routing.yml
```diff
@@ -45,3 +45,11 @@ embed_button.delete:
     _title: 'Delete Embed button'
   requirements:
     _permission: 'administer embed buttons'
+
+entity_embed.autocomplete_entity:
+  path: '/entity-embed/autocomplete/{entity_type_id}'
+  defaults:
+    _controller: '\Drupal\entity_embed\Controller\AutocompleteController::autocompleteEntity'
+  requirements:
+    # @todo What to do here?
```
slashrsm: Good question.... We definitely do not want to expose this to everyone on the internet? Should we dynamically generate permissions for every embed button that is created and check that?

davereid: Hrm. Yeah this is tricky. I'd rather us create a new permission for the autocomplete access itself since it's a crap-shoot trying to rely on the correct permission for each entity type.

Berdir: Yeah, core doesn't have a list access operation. For the dialog, we rely on access to the filter format. I was considering to use the same here, then we could check if that entity type is enabled for that filter. Or we could actually implement access checks on the buttons by falling back on the filter formats/editors that have them enabled?

slashrsm: > Or we could actually implement access checks on the buttons by falling back on the filter formats/editors that have them enabled?

Not sure if I understand what you mean with this. Using simiar approach that we use for dialog sounds ok, since autocomplete always comes in package with it. :)

Another thing crossed my mind. I guess we should also perform some kind of access check against entities before returning them?

cs-shadow:
I also agree with @slashrsm that the route to autocomplete should also work similar to the one for dialog. Access checks on entities would be good to have since it'll be a bad UX if user selects an entity from autocomplete and the embed step rejects it as access denied. So, +1 for that.

Berdir:
Entity queries add a {$entity_type}_access tag by default. It will not return entities that you do not have access to. There are a few special cases that core doesn't handle properly (see entity reference selection plugins/hacks), but not sure if we should worry about that.

On access check on the route, yes we can check the filter, but if we do that, we should also check if that format actually has that button enabled. Not sure if the dialog does that right now? Otherwise you could access any button if you have access to at least one route?

And if we do that, we could inverse the pattern, and say, $button->access('view'), and add an entity access control handler that checks access by checking if it is in any filter format/editor that the user can access. Not sure how easy that would be, as we need to query for editors that have this plugin enabled?

cs-shadow:
> Entity queries add a {$entity_type}_access tag by default. It will not return entities that you do not have access to. There are a few special cases that core doesn't handle properly (see entity reference selection plugins/hacks), but not sure if we should worry about that.

OK, I didn't knew  that. Seems fine then.

> On access check on the route, yes we can check the filter, but if we do that, we should also check if that format actually has that button enabled. Not sure if the dialog does that right now? Otherwise you could access any button if you have access to at least one route?

I'm not sure whether the dialog needs it, because it has a button parameter in the route and will give a 404 if the filter format doesn't has that button enabled. In case of the autocomplete also, I think we could pass the button and format as route parameters and check whether user can access the format. That should be enough I suppose.

Berdir:
> I'm not sure whether the dialog needs it, because it has a button parameter in the route and will give a 404 if the filter format doesn't has that button enabled.

I just checked that and I was able to access the form using a filter/button combination that doesn't actually exit. But fixing that seems like a different issue.

cs-shadow:
OK, that's weird because that's not supposed to happen and we also have
test coverage for that. Since I'm not able to reproduce locally right now,
can you please file a bug report with steps to reproduce this.

Regrads,
Chandan Singh
http://chandansingh.net/

Berdir:
Ok, pushed an update that uses filter access.

Based on above, the access check there is now bogus because you can use any combination of format and button.

As described, I think access checks should move to the buttons, but that's a separate issue.

src/Controller/AutocompleteController.php
```diff
@@ -0,0 +1,50 @@
...
+
+    $entity_type_id = $embed_button->getEntityTypeMachineName();
+    $entity_type = \Drupal::entityManager()->getDefinition($entity_type_id);
+
+    // Prevent errors if the entity type has no label key.
+    if (!$entity_type->hasKey('label')) {
+      return new JsonResponse($matches);
+    }
+
+    $ids = \Drupal::entityQuery($entity_type_id)
```
slashrsm: Would it be nicer to inject entity manager?

src/Form/EntityEmbedDialog.php
```diff
@@ -116,6 +116,11 @@ public function buildForm(array $form, FormStateInterface $form_state, FilterFor
           '#type' => 'textfield',
           '#title' => $this->t('Entity ID or UUID'),
```
slashrsm: This title is probably misleading. Should we change it  to "Entity label" and add #description to explain more?

slashrsm:
Created issue for access checks: https://www.drupal.org/node/2362897

src/Form/EntityEmbedDialog.php
Berdir: Yes, I wasn't sure about it. Could also be just label, or we could even do something fancy and get the title from the label field definition, so it would say Title for nodes, and Filename for files.

any suggestions on the description?

slashrsm: That sounds even better. We could also modify autocomplete route to return `[entity:label] ([entity:uuid])`. 

Description could be something like: "Type entity label and pick the right one from autocomplete suggestions." How does that sound?

Berdir: The problem with returning something like that is that we need similar preg_match() logic as entity reference AutocompleteWidget has, but would certainly be easier to understand, yes. (id) should also be fine, as we'll convert it to uuid anyway.

slashrsm: I am also OK if we use current approach (send UUID over) for now. We can open a follow-up and improve that in the future as it is not super high priority. We are expecting most people to use entity_browser anyway.

Berdir:
Updated with an improved label, using injection for entity query factory (entity manager has a method that we can use).

Berdir:
Kept the current UUID behavior, as suggested.

src/Controller/AutocompleteController.php
```diff
@@ -0,0 +1,77 @@
...
+  protected $queryFactory;
+
+  /**
+   * Constructs an AutocompleteController.
+   *
+   * @param \Drupal\Core\Entity\Query\QueryFactory $query_factory
+   *
+   */
+  public function __construct(QueryFactory $query_factory) {
+    $this->queryFactory =$query_factory;
```
slashrsm: Missing whitespace.

Berdir:
Fixed the whitespace and missing param documentation.

slashrsm:
Thanks! Great UX improvement.

PR #103, opened 2014-10-23
Previewing an entity in the WYSIWYG does not add any attached styles or javascript from the AJAX callback

yched:
PR for https://www.drupal.org/node/2350463

This uses the ajax framework to fetch the previewed entities in the wysiwig. Code is adapted from the work I did on scald D7 for a client (https://www.drupal.org/node/2136415). 

This allows:
- Adding the assets (JS / CSS) for the rendered entity. Works beautifully when the CKeditor is in divarea mode. In iframe mode, the CSS do not get applied to the content of the iframe, but that is to be expected
- Attaching JS behaviors (e.g a picturefill, an image lazy-loading script...). Works fine as well, even in iframe mode, because Drupal.attachBehaviors() is called on an element in the iframe, so `$(selector, context)` works as expected.
  (`$(selector)` runs from the parent document and cannot see the content of the iframe, 
  but `$(selector, SomeElementInTheIframe)` works as expected)

Works a little too well, actually : 
- embedding a node triggers the "contextual links" behavior, which, when unstyled (in the iframe mode D8 uses by default), looks weird.
- embedding a node in a view mode that displays the comment form (!) triggers JS errors in editor.js, because its attach/detach code deesn't fully pass the context to its jquery selectors. Will open a core issue for that.

[edit: those issues have been fixed in latest code - see [comment below](https://github.com/drupal-media/entity_embed/pull/103#issuecomment-60486887)]

### DIFF
--- a/src/Ajax/EntityEmbedInsertCommand.php
+++ b/src/Ajax/EntityEmbedInsertCommand.php
@@ -2,10 +2,10 @@
 
 /**
  * @file
- * Contains \Drupal\entity_embed\EntityEmbedInsertCommand.
+ * Contains \Drupal\entity_embed\Ajax\EntityEmbedInsertCommand.
  */
 
-namespace Drupal\entity_embed;
+namespace Drupal\entity_embed\Ajax;
 
 use Drupal\Core\Ajax\CommandInterface;
 
--- a/src/EntityEmbedInsertCommand.php
+++ b/src/EntityEmbedInsertCommand.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\entity_embed\EntityEmbedInsertCommand.
+ */
+
+namespace Drupal\entity_embed;
+
+use Drupal\Core\Ajax\CommandInterface;
+
+/**
+ * AJAX command for inserting an embedded entity in a CKEditor.
+ *
+ * @ingroup ajax
+ */
+class EntityEmbedInsertCommand implements CommandInterface {
+
+  /**
+   * The HTML content that will replace the matched element(s).
+   *
+   * @var string
+   */
+  protected $html;
+
+  /**
+   * Constructs an EntityEmbedCommand object.
+   *
+   * @param string $html
+   *   String of HTML to be inserted.
+   */
+  public function __construct($html) {
+    $this->html = $html;
+  }
+
+  /**
+   * Implements Drupal\Core\Ajax\CommandInterface:render().
+   */
+  public function render() {
+    return array(
+      'command' => 'entity_embed_insert',
+      'html' => $this->html,
+    );
+  }
+
+}
--- a/js/plugins/drupalentity/plugin.js
+++ b/js/plugins/drupalentity/plugin.js
@@ -87,32 +87,31 @@
         // Fetch the rendered entity.
         init: function () {
           var element = this.element;
-          // Use a throwaway Drupal.ajax object to fetch the HTML, so that we
-          // can retrieve out-of-band assets (JS, CSS...) and attach behaviors.
-          // This requires attaching to an element with a known HTML ID, though.
-          // For now, sticking on the admin_toolbar.
-          // @todo Can we use something else ? Generate an ID on the fly and
-          // assign it to the element itself ?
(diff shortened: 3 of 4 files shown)


yched:
Filed the issue about editor.js : https://www.drupal.org/node/2362637

src/EntityEmbedInsertCommand.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\entity_embed\EntityEmbedInsertCommand.
+ */
+
+namespace Drupal\entity_embed;
```
slashrsm: Shoud we namespace this `Drupal\entity_embed\Ajax`. It looks like a standard pattern in core.

slashrsm:
Tested it and it works as expected. Went through code and didn't notice anything obvious. However, note that I'm not a JS expert.

Thank you!

wimleers:
First: wow!

Second: I'm not sure we want to `Drupal.attachBehaviors()` inside a WYSIWYG editor `<iframe>`?

See discussion at https://www.drupal.org/node/2362637.

yched:
Thanks for the feedback folks :-)

@slashrsm - Ajax namespace : agreed. I hate /src folders being a dumping ground for 10s of unrelated classes. Pushed.

@wimleers : replied in https://www.drupal.org/node/2362637

yched:
Resolved the pending @todo about which DOM element to attach the Drupal.ajax to.
We use the embed element itself, by dynamically assigning it an arbitrary html ID (that gets removed on downcast)

src/Ajax/EntityEmbedInsertCommand.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Drupal\entity_embed\Ajax;
+
+use Drupal\Core\Ajax\CommandInterface;
+
+/**
+ * AJAX command for inserting an embedded entity in a CKEditor.
+ *
+ * @ingroup ajax
+ */
+class EntityEmbedInsertCommand implements CommandInterface {
```
cs-shadow, line 17: Wouldn't it be better to change the name of the class to maybe EntityEmbedInsert (or something similar)? Basically what I want to say is that the word 'Command' is redundant here in the name of the class.

cs-shadow:
Looks good otherwise. Awesome work!

yched:
@cs-shadow : All "ajax command" classes in core are XxxCommand, this is consistent.

slashrsm:
Are test fails related to this PR or we need to chase head?

yched:
Leave out a couple behaviors, as suggested by @wimleers in https://www.drupal.org/node/2362637#comment-9279651

This removes:
- the weird display of node contextual links
- the JS error in editor.js / CK inception when embedding a view mode displaying a comment form.
  
  [note: on current core HEAD, entity_emebed doesn't display comment forms anyway, because cache render placeholders are apparently not expanded anymore, but that is a bug in entity_embed.
  Was broken by the recent 2d3266e - [Issue #2361681 by Wim Leers: drupal_render(): invert second argument ($is_recursive_call -> $is_root_call) => more strict, better DX/TX.](http://www.drupal.org/node/2361681) ]

Currently that list of "unwanted behaviors" is hardcoded. As a followup, it might be interesting to let 3rd party code alter that list, Not sure of the best way to do that though.

I think I'm done here :-)

yched:
@slashrsm : right, just looked into that. The travis report says EntityEmbedHooksTest fails, but it fails on 8.x-1.x too, so yeah, probably a case of HEAD-chasing.

Might be related to the placeholders thing I mentioned above ? But yeah, the core rendering pipe is in flux these days. @wimleers ;-)

cs-shadow:
@yched sorry about that. My bad.

yched:
Cosmetic change : the HTML ID is removed in downcast, so add it in upcast for symmetry.

slashrsm:
Follow-up to fix comment forms: https://www.drupal.org/node/2367479

Also noticed https://www.drupal.org/node/2367483. Probably unrelated  though.

slashrsm:
OK. I think we can merge this. 

Also created follow-up to make unwanted behaviours configurable: https://www.drupal.org/node/2367487

PR #105, opened 2014-11-12
Use recursive rendering

cs-shadow:
d.o issue: https://www.drupal.org/node/2367479


wimleers:
Indeed, as of https://www.drupal.org/node/2361681, you'll want to switch to `drupal_render_root()`.

Berdir:
This should work again as expected once https://www.drupal.org/node/2382667 is fixed.

wimleers:
https://www.drupal.org/node/2382667 has now been committed; this change should no longer be necessary, it should all be working fine with the current `entity_embed` code against 8.x HEAD! :)

PR #107, opened 2014-11-28
Update README, explain Filter Settings

webflo:
The `<drupal-entity>` must be allowed in other all other filter settings.


slashrsm:
Merged. Thanks!

PR #108, opened 2014-12-02
Lint plugin.js

webflo:
I had some issues with `plugin.js` in IE 8 and 9 becuase IE is picky about trailing commas and semicolons.


slashrsm:
Thanks!

PR #112, opened 2014-12-22
Base access checks on embed buttons instead of filter formats

cs-shadow:
d.o issue: https://www.drupal.org/node/2362897

### DIFF
--- a/src/EmbedButtonAccessControlHandler.php
+++ b/src/EmbedButtonAccessControlHandler.php
@@ -26,9 +26,7 @@ class EmbedButtonAccessControlHandler extends EntityAccessControlHandler {
     if ($operation == 'use') {
       return AccessResult::allowedIfHasPermission($account, $embed_button->getPermissionName());
     }
-
-    // No opinion.
-    return AccessResult::neutral();
+    return parent::checkAccess($embed_button, $operation, $langcode, $account);
   }
 
 }
--- a/src/EmbedButtonAccessControlHandler.php
+++ b/src/EmbedButtonAccessControlHandler.php
@@ -1,0 +1,34 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\entity_embed\EmbedButtonAccessControlHandler.
+ */
+
+namespace Drupal\entity_embed;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Entity\EntityAccessControlHandler;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Session\AccountInterface;
+
+/**
+ * Defines the access control handler for the embed button entity type.
+ *
+ * @see \Drupal\entity_embed\Entity\EmbedButton
+ */
+class EmbedButtonAccessControlHandler extends EntityAccessControlHandler {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function checkAccess(EntityInterface $embed_button, $operation, $langcode, AccountInterface $account) {
+    if ($operation == 'use') {
+      return AccessResult::allowedIfHasPermission($account, $embed_button->getPermissionName());
+    }
+
+    // No opinion.
+    return AccessResult::neutral();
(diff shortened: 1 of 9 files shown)


src/EmbedButtonAccessControlHandler.php
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function checkAccess(EntityInterface $embed_button, $operation, $langcode, AccountInterface $account) {
+    if ($operation == 'use') {
+      return AccessResult::allowedIfHasPermission($account, $embed_button->getPermissionName());
+    }
+
+    // No opinion.
```
Berdir: You should call parent::checkAccess() instead, right now, you prevent the admin permission from working, not sure why this is not failing any tests.

src/Entity/EmbedButton.php
```diff
@@ -109,4 +110,11 @@ public function getButtonImage() {
       return file_create_url(drupal_get_path('module', 'entity_embed') . '/js/plugins/drupalentity/entity.png');
     }
   }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getPermissionName() {
```
Berdir, line 117: Maybe include "entity embed" or so in the permission name, to make it a bit clearer to which module it belongs (and prevent clashes)

Berdir:
Looks good in general.

I suppose there is no mechanism to have access control on the displayed buttons?

src/EmbedButtonAccessControlHandler.php
cs-shadow: @Berdir it works because currently we don't use individual admin permissions, we just have a single `Administer embed buttons` which is used. So, actually do not use any other permissions other than `use`. I could change it to `parent::checkAccess()` but in current situation, it'll never be used anyway.

cs-shadow:
@Berdir, I don't think its possible unless ckeditorPlugin introduces a sort of `getPermissions()` function. ~~Or we could implement `hook_ckeditor_plugin_info_alter()`.
Does the latter sound like a good idea?~~

That hook won't work as its for CKEditor plugins and we'll need something for the buttons.

src/Entity/EmbedButton.php
cs-shadow, line 117: I don't think any module in Core does that but if we were to do it, it would be something like `entity embed: use embed button ...`, right?

Berdir, line 117: core modules don't prefix, but they usually have the module name in the permission one way or another. I just meant "use entity embed button $id". There might be other embed modules :)

src/EmbedButtonAccessControlHandler.php
Berdir: Yes, but the single admin permission _is_ checked in the parent/default implementation of checkAccess(). You are not calling that anymore and I don't know why you're not having any test fails... I guess you're not actually using it anywhere, the edit/add routes for example use _permission, not _entity_access.

src/Entity/EmbedButton.php
cs-shadow, line 117: Got it. Will do that.

src/EmbedButtonAccessControlHandler.php
cs-shadow: Tests are not failing because we're not actually using `admin` permissions anywhere. Since for add/edit routes, _permission is used and not _entity_access, I don't think this needs to be changed.

The admin permissions should be required only when we use something like `embed_button.admin` for a more fine grain permissions and not as of now. Am I correct here?

Berdir: No, the admin permission is used for any operation as fallback. You should change _permission to _entity_access: 'entity_embed.edit' and then it will be used (not here), and you must call the parent (here), not doing it is wrong :)

cs-shadow: OK, Done.

PR #113, opened 2014-12-25
Make unwanted behaviors configurable

cs-shadow:
d.o issue: https://www.drupal.org/node/2367487


yched:
But yeah, other that that, looks good enough to me ?

cs-shadow:
@yched Thanks! I'll do that, My knowledge of JS standards is still pretty novice.

PR #114, opened 2014-12-27
Link entity label to its URL in embed form

cs-shadow:
d.o issue: https://www.drupal.org/node/2398975

### DIFF
--- a/src/Form/EntityEmbedDialog.php
+++ b/src/Form/EntityEmbedDialog.php
@@ -161,7 +161,7 @@ class EntityEmbedDialog extends FormBase {
         $form['entity'] = array(
           '#type' => 'item',
           '#title' => $this->t('Selected entity'),
-          '#markup' => $entity->label(),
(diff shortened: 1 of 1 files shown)


src/Form/EntityEmbedDialog.php
```diff
@@ -161,7 +161,7 @@ public function buildForm(array $form, FormStateInterface $form_state, FilterFor
         $form['entity'] = array(
           '#type' => 'item',
           '#title' => $this->t('Selected entity'),
-          '#markup' => $entity->label(),
+          '#markup' => '<a href="' . $entity->url() . '">' . $entity->label() . '</a>',
```
jibran: Why can't we use `l` here?

cs-shadow: @jibran Used `link()` function instead. That's much better.

jibran: `l` is gone anyway :D RTBC++

PR #115, opened 2014-12-27
Use UUID for file icon reference and declare it as dependency

cs-shadow:
d.o issue: https://www.drupal.org/node/2362095

### DIFF
--- a/src/Entity/EmbedButton.php
+++ b/src/Entity/EmbedButton.php
@@ -102,7 +102,7 @@ class EmbedButton extends ConfigEntityBase implements EmbedButtonInterface {
    */
   public function getButtonImage() {
     if ($this->button_icon_uuid) {
-      $image = entity_load_by_uuid('file', $this->button_icon_uuid);
+      $image = $this->entityManager()->loadEntityByUuid('file', $this->button_icon_uuid);
       return $image->url();
     }
     else {
@@ -117,7 +117,7 @@ class EmbedButton extends ConfigEntityBase implements EmbedButtonInterface {
     parent::calculateDependencies();
 
     // Add the file icon entity as dependency if an UUID was specified.
-    if ($this->button_icon_uuid && $file_icon = entity_load_by_uuid('file', $this->button_icon_uuid)) {
+    if ($this->button_icon_uuid && $file_icon = $this->entityManager()->loadEntityByUuid('file', $this->button_icon_uuid)) {
       $this->addDependency($file_icon->getConfigDependencyKey(), $file_icon->getConfigDependencyName());
     }
 
--- a/src/Entity/EmbedButton.php
+++ b/src/Entity/EmbedButton.php
@@ -69,11 +69,11 @@ class EmbedButton extends ConfigEntityBase implements EmbedButtonInterface {
   public $entity_type;
 
   /**
-   * File id of the button's icon.
+   * UUID of the button's icon file.
    *
    * @var string
    */
-  public $button_icon_fid;
+  public $button_icon_uuid;
 
   /**
    * {@inheritdoc}
@@ -101,8 +101,8 @@ class EmbedButton extends ConfigEntityBase implements EmbedButtonInterface {
    * {@inheritdoc}
    */
   public function getButtonImage() {
-    if ($this->button_icon_fid) {
-      $image = file_load($this->button_icon_fid);
+    if ($this->button_icon_uuid) {
+      $image = entity_load_by_uuid('file', $this->button_icon_uuid);
       return $image->url();
     }
     else {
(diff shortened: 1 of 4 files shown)


src/Form/EmbedButtonForm.php
```diff
@@ -98,7 +98,7 @@ public function form(array $form, FormStateInterface $form_state) {
       '#description' => $this->t("Label for the button to be shown in CKEditor toolbar."),
       '#required' => TRUE,
     );
-    $form['button_icon_fid'] = array(
+    $form['button_icon_uuid'] = array(
```
Berdir: Not sure if the form element should be named uuid (or fid), it's a whole file (the value) not just an ID.

src/Form/EmbedButtonForm.php
```diff
@@ -119,7 +119,13 @@ public function form(array $form, FormStateInterface $form_state) {
   public function save(array $form, FormStateInterface $form_state) {
     $embed_button = $this->entity;
 
-    $embed_button->button_icon_fid = count($embed_button->button_icon_fid) ? $embed_button->button_icon_fid[0] : NULL;
+    // $embed_button->button_icon_uuid = count($embed_button->button_icon_uuid) ? $embed_button->button_icon_uuid[0] : NULL;
+    if (count($embed_button->button_icon_uuid) && $file = entity_load('file', $embed_button->button_icon_uuid[0])) {
```
Berdir: count() ?

I would suggest to move this logic to copyFormValuesToEntity(), then you can get the file directly from $form_state values, and assign it to the entity. Renaming the form element would also make this part here less confusing I think.

Berdir:
Added two remarks, but in general, this looks good to me.

src/Form/EmbedButtonForm.php
davereid: Agreed, let's normalize this to just 'button_icon' since it's a managed file element.

src/Entity/EmbedButton.php
```diff
@@ -101,12 +101,27 @@ public function getButtonLabel() {
    * {@inheritdoc}
    */
   public function getButtonImage() {
-    if ($this->button_icon_fid) {
-      $image = file_load($this->button_icon_fid);
+    if ($this->button_icon_uuid) {
+      $image = entity_load_by_uuid('file', $this->button_icon_uuid);
```
jibran: `entity_load_by_uuid` can be replaced by `$this->entityManager()->loadEntityByUuid()`.

src/Form/EmbedButtonForm.php
```diff
@@ -61,7 +61,7 @@ public function form(array $form, FormStateInterface $form_state) {
     $form = parent::form($form, $form_state);
 
     $embed_button = $this->entity;
-    $button_image = $embed_button->button_icon_fid ? array($embed_button->button_icon_fid) : array();
+    $button_image = $embed_button->button_icon_uuid ? array($embed_button->button_icon_uuid) : array();
     $file_scheme = \Drupal::config('entity_embed.settings')->get('file_scheme');
```
jibran, line 73: `\Drupal::config('entity_embed.settings')` can be injected using constructor. Maybe follow up.

src/Form/EmbedButtonForm.php
jibran: `entity_load()` can be replaced by `$this->entityManager->getStorage('file')->load()` or `File::load()`

src/Form/EmbedButtonForm.php
cs-shadow: Renamed to `button_icon`.

src/Form/EmbedButtonForm.php
cs-shadow: Done and done.

src/Form/EmbedButtonForm.php
cs-shadow, line 73: OK, I'll open a follow-up for this after this PR gets in to avoid conflicts.

slashrsm:
Looks good but needs reroll.

src/Form/EmbedButtonForm.php
```diff
@@ -61,7 +60,16 @@ public function form(array $form, FormStateInterface $form_state) {
     $form = parent::form($form, $form_state);
 
     $embed_button = $this->entity;
-    $button_image = $embed_button->button_icon_fid ? array($embed_button->button_icon_fid) : array();
+
+    // Get default for button image. If its uuid is set, get the id of the file
+    // to be used as default in the form.
+    $button_icon = array();
+    if ($embed_button->button_icon_uuid) {
+      $files = $this->entityManager->getStorage('file')->loadByProperties(array('uuid' => $embed_button->button_icon_uuid));
```
davereid, line 68: This could have also used $this->loadEntity('file', $embed_button->button_icon_uuid) since the method is available via EntityHelperTrait

davereid, line 68: Or at least $this->entityManager()->loadEntityByUuid('file', $this->button_icon_uuid) like above in calculateDependencies()

cs-shadow, line 68: @davereid https://github.com/drupal-media/entity_embed/pull/116

PR #118, opened 2015-01-07
Use custom access check to ensure that text format has embed button enabled

cs-shadow:
d.o issue: https://www.drupal.org/node/2362897

This PR does two things:
1. Reverts previous merge which based access checks on embed buttons and introduced new permissions.
2. Adds a new `buttonIsEnabled()` function which takes a filter format and an embed button and returns whether or not the embed button is enabled for given text format. The same function is used for custom access check on the routes.


Berdir:
entity_embed right now seems broken because it is trying to access $filter_format which was removed, so it fatals. So it would be great to get that in asap.

Only thing that I can see is that we are using the button label to identify the embed button, is there nothing else available? That thing needs to have some sort of ID?

PR #120, opened 2015-01-26
Added an alter hook to allow modules to influence available options

Berdir:


cs-shadow:
Can't we use `hook_entity_embed_display_plugins_alter` which is already available?

davereid:
I like the idea of this. Maybe it would be more useful to have this alter call up in EntityEmbedDisplayManager::getDefinitionOptionsForEntity() instead, rather than just the form builder?

Berdir:
hook_entity_embed_display_plugins_alter is the plugin (derivate) hook, that is global, this hook is for a specific entity type. But I'm totally open for better names, something that better reflects that, but a good name that is not too long might not be easy to find. We could add a _for_entity to that, to be consistent with the method?

I'm OK with doing it directly in getDefinitionOptionsForEntity(), I originally thought we'd need more context from the form, but as long as we keep the default_value check there then we should be fine.

Berdir:
Ok, updated, renamed to hook to hook_entity_embed_display_plugins_for_context_alter() and moved into getDefinitionsForContexts() and added some documentation.

cs-shadow:
@davereid Looks good to me.

slashrsm:
I think we can merge this. Looks good to me too and it seems that we agree on the approach.

PR #121, opened 2015-01-30
Preview entity link in a new window

cs-shadow:
d.o issue: https://www.drupal.org/node/2417067


slashrsm:
Looks good to me, but I'd really like someone to give another look at JS part at least.

cs-shadow:
Well, FWIW the JS part is mostly duplicated from: `Drupal.behaviors.filePreviewLinks` which does the same thing for files.

slashrsm:
OK. Let's merge.

PR #122, opened 2015-01-30
Filter out entity types which do not have a display plugin from button configuration form

cs-shadow:
d.o issue: https://www.drupal.org/node/2417105


src/Form/EmbedButtonForm.php
```diff
@@ -73,6 +73,23 @@ public function form(array $form, FormStateInterface $form_state) {
     $upload_directory = \Drupal::config('entity_embed.settings')->get('upload_directory');
     $upload_location = $file_scheme . '://' . $upload_directory . '/';
 
+    $entity_types = $this->entityManager->getEntityTypeLabels(TRUE);
+    $filtered_entity_types = array();
+    // Add all Content entites by default.
+    $filtered_entity_types['Content'] = $entity_types['Content'];
```
Berdir: Those groups are translatable strings, so that won't work. Maybe rely on the currently official but complicated method with isSubclassOf, see https://www.drupal.org/node/2194783?

And it might be easier to just build the list yourself with getDefinitions(), then you can for example also call $entity_type->hasViewBuilder().

src/Form/EmbedButtonForm.php
```diff
@@ -73,6 +73,23 @@ public function form(array $form, FormStateInterface $form_state) {
...
+    // Select only those config entities which have a view builder.
+    $filtered_config_entities = array();
+    foreach ($entity_types['Configuration'] as $entity_type => $label) {
+      if ($this->entityManager->hasHandler($entity_type, 'view_builder')) {
```
Berdir: Also, i can actually imagine use cases where it makes sense to embed config entities despite not having a view builder, through custom field formatters. We are using custom formatters for contact forms with entity references, and I can imagine the same for views (entity embed for views would be pretty cool ;))

cs-shadow: What do you suggest should be the default behavior then? Do you have any other (more suitable) criteria for dis-allowing config entities?

davereid: I would say contact forms in core should have a view builder that displays the form, not needing to support the other way around.

src/Form/EmbedButtonForm.php
```diff
@@ -168,6 +168,39 @@ protected function copyFormValuesToEntity(EntityInterface $entity, array $form,
...
+   * @return array
+   *   An array of entity type labels, keyed by entity type name.
+   */
+  protected function getFilteredEntityTypes() {
+    $options = array();
+    $definitions = $this->entityManager->getDefinitions();
+
+    foreach ($definitions as $entity_type_id => $definition) {
+      // Don't include configuration entities which do not have a view builder.
+      if ($definition->getGroup() != 'configuration' || $definition->hasViewBuilderClass()) {
```
davereid, line 185: Could we simplify this to just `$definition->hasViewBuilderClass()`? Or maybe I don't understand why this is an OR instead of an AND condition.

cs-shadow, line 185: I thought that it might be confusing. The idea is following:
- Config entities with a view builder are allowed.
- All content entities, even without a view builder (like files).

So, based on that all those entities are allowed which are either "not config" or have a view builder. Hence, its an OR instead of AND.

Do you think it'll be more clear if I wrote `if ($definition->getGroup() == 'content' || $definition->hasViewBuilderClass()` instead?

davereid, line 185: Hrm. @cs-shadow Maybe another way thinking about this is once we have EntityDisplayPluginManager::getDefinitionsForEntityType() (from https://www.drupal.org/node/2349371), we could see which entity types have display plugins, and use that to filter the list, instead of this logic.

cs-shadow, line 185: Agreed, that'll be cleaner. Will first add that method first in a separate PR

cs-shadow, line 185: @davereid I'm not sure how this is going to work since there'll be some foramtters whose `entity_types` value is FALSE and hence are applicable to all entity types. So, no matter which entity type we choose, they'll always be present. This again leaves us with two options:
- Use the same `hasViewBuilder` approach, or
- Check whether that entity type has any formatter besides default ones

What do you think?

cs-shadow:
@davereid A snapshot of select list:

screenshot from 2015-02-12 23 07 40

src/Form/EmbedButtonForm.php
slashrsm, line 185: I'll go ahead and merge this as it improves existing behaviour. If we decide to go with the other approach we can do so in a follow-up.

PR #123, opened 2015-01-30
Use EntityDeleteForm to handle entity deletion

cs-shadow:
d.o issue: https://www.drupal.org/node/1728804


src/Form/EmbedButtonDeleteForm.php
```diff
@@ -30,20 +22,4 @@ public function getCancelUrl() {
     return new Url('embed_button.list');
```
Berdir: If you use the new collection link template, then you can remove this part too.

cs-shadow: If I remove this function also, the file will essentially be empty. In that case, can I directly use the EntityDeleteForm as handler for delete?

Berdir: Yes, you can :)

cs-shadow: Thats beautiful, great work @berdir on that issue.

On Fri, Jan 30, 2015 at 6:30 PM, Sascha Grossenbacher <
notifications@github.com> wrote:

> In src/Form/EmbedButtonDeleteForm.php
> https://github.com/drupal-media/entity_embed/pull/123#discussion_r23840573
> :
> 
> > @@ -30,20 +22,4 @@ public function getCancelUrl() {
> >      return new Url('embed_button.list');
> 
> Yes, you can :)
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/drupal-media/entity_embed/pull/123/files#r23840573.

## 

Regrads,
Chandan Singh
http://chandansingh.net/

davereid:
Yay, great cleanup!

PR #124, opened 2015-01-31
Attch resources automatically while rendering forms in Ajax Response

cs-shadow:
d.o issue: https://www.drupal.org/node/2414557


Berdir:
Tested this, works great. +1 to merging it. Already using in my project.

roidanton:
Furthermore this fix removes drupal_render() which is [marked as deprecated](https://api.drupal.org/api/drupal/core%21includes%21common.inc/function/drupal_render/8).

slashrsm:
Thanks!

PR #125, opened 2015-01-31
Change ajax auto completion to CONTAINS instead of STARTS_WITH

cs-shadow:
Seems like a minor UX improvement to me.


Berdir:
Not sure this is really better, if you have a lot of content, then you often get a lot of unrelated matches with CONTAINS. STARTS_WITH is usually what you expect as a user, no?

cs-shadow:
I'm not sure what's best here either. CONTAINS works better for me personally but your argument about getting a lot of unrelated also seems legit. What I thought was that if you had a piece of content "Foo Bar", you might search it by "Bar" instead of "Foo".

Maybe, this can be made configurable?

slashrsm:
Another thing is performance. CONTAINS can be much slower than STARTS_WITH, right? You'd definitely notice that on a site with a huge amount of content. 

Definitely +1 for making it configurable if we go down that route.

cs-shadow:
Okay, closing this one for now. Will open a new one which allows it to be configurable if we decide to go along that path.

PR #127, opened 2015-02-09
Add getDefinitionsForEntityType() method to EntityDisplayPluginManager

cs-shadow:
d.o issue: https://www.drupal.org/node/2423113

### DIFF
--- a/src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
+++ b/src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
@@ -83,7 +83,7 @@ class EntityEmbedDisplayManager extends DefaultPluginManager {
   }
 
   /**
-   * Provides an list of plugins that can be used for a certain entity.
+   * Provides a list of plugins that can be used for a certain entity.
    *
    * @param \Drupal\Core\Entity\EntityInterface $entity
    *   An entity object.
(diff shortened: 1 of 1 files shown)


src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
```diff
@@ -98,4 +98,24 @@ public function getDefinitionOptionsForEntity(EntityInterface $entity) {
...
+   *   The entity type id.
+   *
+   * @return array
+   *   An array of valid plugin labels, keyed by plugin ID.
+   */
+  public function getDefinitionOptionsForEntityType($entity_type) {
+    $definitions = $this->getDefinitions();
+    $valid_definitions = array();
+    foreach ($definitions as $id => $definition) {
+      if ($definition['entity_types'] == FALSE || in_array($entity_type, $definition['entity_types'])) {
```
slashrsm, line 114: empty() maybe? It is more general and covers == FALSE too.

src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
```diff
@@ -98,4 +98,24 @@ public function getDefinitionOptionsForEntity(EntityInterface $entity) {
...
+        $valid_definitions[$id] = $definition['label'];
```
slashrsm, line 115: Would it make sense to return entire definitions?

src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
cs-shadow, line 114: I don't think that'll work. `FALSE` is special in the sense that if entity_types is False then it means that its applicable to all entity types. Changing it to `empty()` would imply that it applies to no entity type which is not what we want here.

src/EntityEmbedDisplay/EntityEmbedDisplayManager.php
cs-shadow, line 115: No really. Since this function will mostly be used to populate options in form, its better to just return the label otherwise the caller will have to extract the label from the definition.

FWIW, this behavior is similar to the existing `getDefinitionOptionsForEntity()` method.

slashrsm:
OK. Let's merge then.

PR #129, opened 2015-02-15
Refactor access check logic of display plugins

cs-shadow:
d.o issue: https://www.drupal.org/node/2425427


src/EntityEmbedDisplay/FieldFormatterEntityEmbedDisplayBase.php
```diff
@@ -7,13 +7,53 @@
...
 
   /**
+   * The field formatter plugin manager.
+   *
+   * @var \Drupal\Core\Field\FormatterPluginManager
+   */
+  protected $formatterPluginManager;
+
+  /**
+   * {@inheritdoc}
```
slashrsm: Probably needs something more verbose.

PR #130, opened 2015-03-04
Use lower case and underscores instead of camel casing for variable names

prateekmehta:
drupal.org issue link: https://www.drupal.org/node/2445737


cs-shadow:
Merged. Thanks!

PR #132, opened 2015-03-08
Make alt attribute required for image field formatter form

prateekmehta:
d.o issue: https://www.drupal.org/node/2417525
D8 core issue: https://www.drupal.org/node/2303765

### DIFF
--- a/src/Plugin/entity_embed/EntityEmbedDisplay/ImageFieldFormatter.php
+++ b/src/Plugin/entity_embed/EntityEmbedDisplay/ImageFieldFormatter.php
@@ -79,6 +79,7 @@ class ImageFieldFormatter extends FileFieldFormatter {
       '#description' => $this->t('This text will be used by screen readers, search engines, or when the image cannot be loaded.'),
       '#parents' => array('attributes', 'alt'),
       // @see http://www.gawds.org/show.php?contentid=28
+      '#required' => TRUE,
       '#maxlength' => 512,
     );
     $form['title'] = array(


cs-shadow:
Looks good to me apart from the fact that Core gives an option to opt-out of the required alt text. I'm not entirely sure whether or not we should do that too. Maybe @slashrsm or @davereid can comment on that.

davereid:
Yes, I'd definitely like to support the opt-out.

cs-shadow:
@davereid Yes, I guessed so. But I'm not sure where should that option go since we don't have a system currently which allows to configure settings for a particular plugin. So do you want a global module-level setting or a new provision for plugin-specific settings?

davereid:
Hrm, I don't think we need to add any settings per-say. We just need to support the user entering `""` in the alt text as an empty value. Or am I over-simplifying it?

davereid:
Hrm, so Drupal 8 add it as a setting on the field itself. _thinking_

cs-shadow:
@davereid There are two things to keep in mind here:
1. Its possible to make alt text not required in Core by changing field settings. I guess we can get the config from image field setting and use that instead of adding our own config.
2. Even if the alt text is required, end users can enter `""` and thats treated as empty alt text. We'd want this too but from what it seems this would require some JS hacks.

kattekrab:
Seems the behaviour is different for images added via CKeditor vs images added via an image field.  CKeditor prompts the user to add "" if they want to include empty/null alt text.   The issue alt text to be required doesn't include this behaviour, but I reckon it would be a nice follow up if it did - and if entity embed did too, we'd have a consistency win, and a big accessibility win. 

Thanks for pulling this in to contrib.  I'm really excited by this! :)

cs-shadow:
@kattekrab Yes, we plan to make it as consistent as possible with the Core image dialog. I'll close this PR and open a new one where the functionality is similar to the Core dialog

cs-shadow:
Closing in favor of #136.

PR #134, opened 2015-03-08
Ensure 'data-view-mode' and 'data-entity-embed-display' is present during upcast

yuvraj03:
drupal.org issue
https://www.drupal.org/node/2448411


cs-shadow:
Merged, thanks!

PR #135, opened 2015-03-13
Configure CKEditor DTD for drupal-entity element

cs-shadow:
d.o issue: https://www.drupal.org/node/2448449

That issue has been moved to Core as it might be a bug, but it was pointed out that we must configure DTD any way. This PR fixes the issue with our CKEditor plugin.

### DIFF
--- a/js/plugins/drupalentity/plugin.js
+++ b/js/plugins/drupalentity/plugin.js
@@ -13,9 +13,22 @@
 
     // The plugin initialization logic goes inside this method.
     beforeInit: function (editor) {
+      // Configure CKEditor DTD for custom drupal-entity element.
+      // @see https://www.drupal.org/node/2448449#comment-9717735
+      var dtd = CKEDITOR.dtd, tagName;
+      dtd['drupal-entity'] = {'#': 1};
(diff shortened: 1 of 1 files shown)


js/plugins/drupalentity/plugin.js
```diff
@@ -13,9 +13,22 @@
...
     // The plugin initialization logic goes inside this method.
     beforeInit: function (editor) {
+      // Configure CKEditor DTD for custom drupal-entity element.
+      // @see https://www.drupal.org/node/2448449#comment-9717735
+      var dtd = CKEDITOR.dtd, tagName;
+      dtd['drupal-entity'] = {'#': 1};
+      // Register drupal-entity element as allowed child, in each tag that can
+      // contain a p element.
+      for (tagName in dtd) {
+        if (dtd[tagName].p) {
```
davereid: p and not div?

cs-shadow: Well, I initially thought that div might not be allowed for all filter formats but on second thoughts it doesn't seem likt dtd would have that info. I guess I'll change it back to div

PR #137, opened 2015-03-23
Use '#type => status_messages' instead of '#theme => status_messages'

cs-shadow:
Chase Core. Original change record: https://www.drupal.org/node/2446615


slashrsm:
Needs reroll.

cs-shadow:
Should be fine now.

slashrsm:
Thanks!

PR #138, opened 2015-03-23
Raise an exception if no display option is available for selected entity

cs-shadow:
d.o issue: https://www.drupal.org/node/2457957

### DIFF
--- a/src/Form/EntityEmbedDialog.php
+++ b/src/Form/EntityEmbedDialog.php
@@ -21,6 +21,7 @@ use Drupal\entity_embed\EntityHelperTrait;
 use Drupal\entity_embed\EmbedButtonInterface;
 use Drupal\filter\FilterFormatInterface;
 use Drupal\Component\Serialization\Json;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Form/EntityEmbedDialog.php
```diff
@@ -306,13 +321,27 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+              // proceeding to the next step. Rasie an error otherwise.
+              $embed_button = $form_state->get('embed_button');
+              $allowed_plugins = $embed_button->getAllowedDisplayPlugins();
+              $available_plugins = $this->displayPluginManager()->getDefinitionOptionsForEntity($entity);
+              $display_plugin_options = empty($allowed_plugins) ? $available_plugins : array_intersect_key($available_plugins, $allowed_plugins);
+              // If no plugin is available after taking the intersection,
+              // raise error. Also log an exception.
+              if (empty($display_plugin_options)) {
+                $form_state->setError($form['attributes']['data-entity-id'], $this->t('No display options available for the selected entity. Please select another entity.'));
+                $this->logger->warning('No display options available for "@type:" entity "@id" while embedding using button "@button". Please ensure that at least one display plugin is allowed for this embed button which is available for this entity.', array('@type' => $entity_type, '@id' => $entity->id(), '@button' => $embed_button->id()));
```
slashrsm, line 343: Do we need " around @type etc.?

cs-shadow, line 343: Thats not required per se but just added them so that the user can distinguish between the actual values and rest of the text.

PR #145, opened 2015-06-28
Fixed call to Drupal.ajax as per https://www.drupal.org/node/2484581.

davereid:


davereid:
I confirmed this fixed the embedding and ajax preview. The ajax preview itself is still broken due to https://www.drupal.org/node/2496363, so that is being fixed separately.

PR #146, opened 2015-07-02
2496363 Fix filter rendering

davereid:


src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -113,25 +107,33 @@ public function process($text, $langcode) {
...
+            $accessMetaData = CacheableMetadata::createFromObject($access);
+            $entityMetadata = CacheableMetadata::createFromObject($entity);
 
-            if ($entity->access('view')) {
+            // @todo Swap these https://www.drupal.org/node/2516802 is fixed in core.
+            $result->addCacheContexts($accessMetaData->getCacheContexts());
+            $result->addCacheTags($accessMetaData->getCacheTags());
+            $result->addCacheContexts($entityMetadata->getCacheContexts());
+            $result->addCacheTags($entityMetadata->getCacheTags());
+            //$result = $result->merge($entityMetadata)->merge($accessMetadata);
```
cs-shadow: Can this line be removed?

src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -113,25 +107,33 @@ public function process($text, $langcode) {
...
-            // Add cache tags and contexts.
-            if ($tags = $entity->getCacheTags()) {
-              $result->addCacheTags($tags);
-            }
-            if ($contexts = $entity->getCacheContexts()) {
-              $result->addCacheContexts($contexts);
-            }
+            $access = $entity->access('view', NULL, TRUE);
+            $accessMetaData = CacheableMetadata::createFromObject($access);
+            $entityMetadata = CacheableMetadata::createFromObject($entity);
```
cs-shadow: nitpick: Should use underscore separated variable names AFAIK.

cs-shadow:
@davereid Also, can you take a look at the failing tests. I suppose some tests need to be changed too now.

src/Plugin/Filter/EntityEmbedFilter.php
davereid: Let's leave it for now. This allows the module to work with D8 at least older than 24 hours at this point. :) We can swap this at the next beta or RC release.

cs-shadow: Sure.

davereid:
@cs-shadow Something I'd love to get feedback on here. Previously, we would leave the wrapping 'div' around the embedded content. On failure, we also would just leave the empty wrapping div with the embed attributes alone. With this patch, we completely replace the wrapping div with the embedded content, and also on failure remove the embedding div. Thoughts?

cs-shadow:
@davereid The idea behind retaining the original `div` was that it could provide a good point of reference for theming - say if someone wants to theme embeds in a particular way he could use that element. Or one could add certain classes, ids or even some other data attributes for another purpose so it seemed nice to retain those things.

In cases when embed is unsuccessful, I try to draw a parallel to a case where an `src` of an `img` or `iframe` tag is corrupt. In those cases while the actual element is retained.

So, I would personally prefer to retain that wrapping `div` (it could be anything other than div too) unless there are some advantages by removing it.

Are there any such advantages which I might have overlooked?

davereid:
@cs-shadow I reverted back to the original behavior, which leaves the embed div in place. I think it might be worth separate discussion.

PR #149, opened 2015-07-15
Fixed $entity->access() cannot be used for public files. Use EntityHelperTrait::accessEntity() instead.

davereid:
Had to tweak EntityHelperTrait::accessEntity() to make it compatible with the return values of $entity->access() and returning AccessResult objects instead of simple booleans.

### DIFF
--- a/src/EntityHelperTrait.php
+++ b/src/EntityHelperTrait.php
@@ -8,6 +8,7 @@
 namespace Drupal\entity_embed;
 
 use Drupal\Component\Utility\SafeMarkup;
+use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityManagerInterface;
 use Drupal\Core\Entity\EntityStorageException;
@@ -198,6 +199,8 @@ trait EntityHelperTrait {
   /**
    * Check access to an entity.
    *
+   * @todo Remove when https://www.drupal.org/node/2533978 is fixed in core.
+   *
    * @param \Drupal\Core\Entity\EntityInterface $entity
    *   The entity object.
    * @param string $op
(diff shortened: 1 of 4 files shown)


Berdir:
Somewhat related core issue: https://www.drupal.org/node/2310307

davereid:
So, trying to resolve the remaining failure, it seems that there is an issue with our setDomNodeContent() function.

$entity_output during FileFieldFormatterTest:

```
<span class="file file--mime-text-plain file--text"> <a href="http://drupal8.dev/sites/simpletest/577875/files/example.txt" type="text/plain; length=8" title="example.txt">This is sample description</a></span>
```

Output HTML during FileFieldFormatterTest:

```
<div data-entity-type="file" data-entity-uuid="450d7516-a364-4e66-ab04-04df4411cc62" data-entity-embed-display="file:file_default" data-entity-embed-settings="{&quot;description&quot;:&quot;This is sample description&quot;}"><span class="file file--mime-text-plain file--text"> <a href="http://drupal8.dev/sites/simpletest/577875/files/example.txt" type="text/plain; length=8" title="example.txt">This is sample description</a></span></div>
```

$entity_output during ImageFieldFormatterTest:

```
<img src="http://drupal8.dev/sites/simpletest/400888/files/example.png" alt="This is sample description" title="This is sample title" />
```

Output HTML during ImageFieldFormatterTest:

```
<div data-entity-type="file" data-entity-uuid="d79db16d-a40d-4fe8-8d61-97ee761d31ad" data-entity-embed-display="image:image" data-entity-embed-settings="{&quot;image_link&quot;:&quot;&quot;}" alt="This is sample description" title="This is sample title">  </div>
```

I'm not sure why image embedding worked previously...

davereid:
The problem is definitely this:

``` php
      // Load the contents into a new DOMDocument and retrieve the element.
      $replacement_node = Html::load($content)->getElementsByTagName('body')
        ->item(0)
        ->childNodes
        ->item(0);

      // Import the updated DOMNode from the new DOMDocument into the original
      // one, importing also the child nodes of the replacement DOMNode.
      $replacement_node = $node->ownerDocument->importNode($replacement_node, TRUE);
```

This ends up munging the image HTML tag to an empty string.

PR #151, opened 2015-07-22
Button icons are deleted after cron

webflo:
PR for https://www.drupal.org/node/2537786

### DIFF
--- a/src/Entity/EmbedButton.php
+++ b/src/Entity/EmbedButton.php
@@ -9,8 +9,10 @@ namespace Drupal\entity_embed\Entity;
 
 use Drupal\Core\Config\Entity\ConfigEntityBase;
 use Drupal\editor\EditorInterface;
+use Drupal\Core\Entity\EntityStorageInterface;
 use Drupal\entity_embed\EmbedButtonInterface;
 use Drupal\entity_embed\EntityHelperTrait;
+use Drupal\file\FileUsage\FileUsageInterface;
 
 /**
  * Defines the EmbedButton entity.
@@ -96,6 +98,13 @@ class EmbedButton extends ConfigEntityBase implements EmbedButtonInterface {
   public $display_plugins;
 
   /**
+   * The file usage service.
+   *
+   * @var \Drupal\file\FileUsage\FileUsageInterface
+   */
+  protected $fileUsage;
+
+  /**
    * {@inheritdoc}
    */
   public function getEntityTypeMachineName() {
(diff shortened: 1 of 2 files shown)


src/Tests/EmbedButtonAdminTest.php
```diff
@@ -99,4 +100,40 @@ public function testEmbedButtonAdmin() {
     $this->assertNoText($name, 'Test embed_button does not appears on the list page');
   }
 
+  /**
+   * Tests the embed_button and file usage integration.
+   */
+  public function testEmbedButtonIcon() {
```
cs-shadow: I think this test will be more suited if it were moved to EmbedButtonCrudTest since we're not really testing administrative interface here.

webflo: Right, but EmbedButtonCrudTest is a KernelTest and the test needs the db schema from file module, entity schema for file and user (because file has a dependency on user). Not sure if it worth the trouble. But sure i can split up the test and add a new testclass.

webflo: @cs-shadow Moved the test to EmbedButtonCrudTest. It was relatively easy ..

davereid:
@webflo Can I have you move the functions from EntityHelperTrait to the EmbedButton class itself? Like on https://github.com/drupal-media/embed/pull/7/files#diff-28f929be92c4e659804666de039e8749R178 ? I don't think anything else will really need to re-use this functionality so it seems odd to put it in the trait.

src/Entity/EmbedButton.php
```diff
@@ -170,4 +171,49 @@ public function calculateDependencies() {
...
+  /**
+   * {@inheritdoc}
+   */
+  public function postSave(EntityStorageInterface $storage, $update = TRUE) {
+    parent::postSave($storage, $update);
+
+    $new_button_icon_uuid = $this->get('button_icon_uuid');
+    if (isset($this->original)) {
+      $unchanged = $this->original;
+      $old_button_icon_uuid = $unchanged->get('button_icon_uuid');
```
davereid: Just make this `$old_button_icon_uuid = $this->original->get('button_icon_uuid')`. There's no need to define `$unchanged` then.

webflo: `$this->original` not defined during the initial entity save.

davereid: Yes, I'm not proposing to remove the `if (isset($this->original)) {` check. I'm just pointing out that the `$unchanged` variable is not used anywhere else, so we should just not need it, and call the get() method directly on `$this->original`.

src/Tests/EmbedButtonCrudTest.php
```diff
@@ -126,4 +127,47 @@ protected function deleteTests() {
...
+      'id' => 'test_button',
+      'label' => 'Testing embed button instance',
+      'button_label' => 'Test',
+      'entity_type' => 'node',
+      'entity_type_bundles' => array('article'),
+      'button_icon_uuid' => $file1->uuid(),
+      'display_plugins' => array('default'),
+    );
+
+    $entity = entity_create('embed_button', $button);
```
davereid: The entity type has changed to 'entity_embed_button' now, and will need to be updated to pass tests.

cs-shadow:
Looks good to me now.

@davereid Anything holding this back?

davereid:
Blargh, sorry more merge conflicts.

webflo:
Fixed the conflicts with git rebase.

davereid:
Thanks @webflo for helping fix this, and for your persistence.

PR #152, opened 2015-07-26
Fix serialization exceptions on form submission

Berdir:
The problem is that DependencySerializationTrait can only deal with services, because only those have the magic _serviceId property that it needs.

Registering the log channel as a service with a factory makes the form work again for me. Let's see if that also helps with fixing some of those test fails.


jibran:
+1 to merge.

Berdir:
Same amount of test fails, I guess there are no tests for the embed UI?

Berdir:
#153 fixed this in a different way.

PR #154, opened 2015-08-05
Require Embed module and start using DomHelperTrait.

davereid:


cs-shadow:
Embed module isn't installed on Travis, seems like something's wrong with composer_manager.

davereid:
Yeah I think the first problem was Embed didn't have a release node on
Drupal.org.

webflo:
Embed module is not listed on https://packagist.org/ and AFAIK composer_manager does not use https://packagist.drupal-composer.org/ by default.

davereid:
No, we shouldn't need to download it with composer, Drush should be doing it for us, but I think we'll still be bitten by the Drush download loop, so we manually have to download it in our drupal_ti script.

src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -143,7 +145,9 @@ public function process($text, $langcode) {
           watchdog_exception('entity_embed', $e);
         }
 
-        $this->setDomNodeContent($node, $entity_output);
+        // Ensure this element is using <div> now instead of <drupal-entity>.
+        $this->changeNodeName($node, 'div');
```
cs-shadow: I suggest changing node name only if the initial tag was `<drupal-entity>`. If the filter was used directly without WYSIWYG it might have wrapped in something other than `div` and we shouldn't change that I suppose.

PS: That is the only failing test in this PR and if we indeed want this behavior instead then we need to change the final test case in `EntityEmbedFilterTest`

cs-shadow:
@davereid I've restored the original behavior - i.e. rename to `div` only if it was `drupal-entity` and not otherwise for the time being to fix the tests. Feel free to change it if you think it should be `div` always.

Also, since `drupal/embed` is not a valid composer package at the moment, do you think its a good idea to add it? If one tries to install it with composer, it won't work so whats the point?

cs-shadow:
This looks good now.

cs-shadow:
@davereid Are you waiting on some other issue to merge this?

davereid:
I was waiting for any feedback since I mentioned this specifically in our scrum call this week. Also wasn't sure if maybe we should tag something before this is merged so all the new embed module dependency isn't required for anyone looking to use the module. But we could always tag later I guess.

PR #155, opened 2015-08-06
Fix conflict with Embed module button configuration entity.

davereid:
This shouldn't prevent any exports from working since this doesn't change the config entity filenames at all. Just prevents the menu paths from conflicting.

### DIFF
--- a/entity_embed.info.yml
+++ b/entity_embed.info.yml
@@ -6,4 +6,4 @@ package: Filters
 dependencies:
   - editor
   - filter
-configure: embed_button.list
+configure: entity_embed_button.list
--- a/entity_embed.links.action.yml
+++ b/entity_embed.links.action.yml
@@ -1,5 +1,5 @@
 entity_embed.embed_button_add:
-  route_name: embed_button.add
+  route_name: entity_embed_button.add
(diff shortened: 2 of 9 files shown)


cs-shadow:
Since we've changed the id of the config entity, we'll need to modify it in the places where its being used as well and that should fix the failing tests.

From a quick glance, we need to fix the tests at least in https://github.com/drupal-media/entity_embed/blob/8.x-1.x/src/Tests/EmbedButtonCrudTest.php#L38, https://github.com/drupal-media/entity_embed/blob/8.x-1.x/src/Plugin/CKEditorPlugin/DrupalEntity.php#L65 and https://github.com/drupal-media/entity_embed/blob/8.x-1.x/src/Form/EmbedButtonForm.php#L322.

davereid:
@cs-shadow Want to give this a look-over again, should resolve everything now. I did a code search for any remaining instances of embed_button that didn't get changed to entity_embed_button.

cs-shadow:
Yep, looks good now. :+1:

jonhattan:
I think this PR introduced a regression. https://www.drupal.org/node/2548987

PR #157, opened 2015-08-08
Add tests to ensure that node config entity is installed by default

cs-shadow:


cs-shadow:
Nevermind, was testing something.

PR #158, opened 2015-08-10
Attempt to fail tests.

davereid:


davereid:
Hmmmm.... EntityEmbedDialogTest is not getting run.

davereid:
Oh, that file is local for me for some reason.

cs-shadow:
@davereid thats because we don't have a EntityEmbedDialogTest (yet). I was blocked by this issue long back: https://www.drupal.org/node/2328659

Will need to tweak a couple of things in the code from that issue if you still want to try that out.

davereid:
Ah, I'm going to propose adding it back in at least for the test coverage for access to the embed form.

cs-shadow:
@davereid Yep, would definitely like that.

PS: Just to clarify, we wouldn't be adding it "back" per se since it was never added in the first place. I'll try see if that form post issue is resolved now.

PR #160, opened 2015-08-10
Scrutinizer fixes

davereid:


davereid:
@cs-shadow Want to give an OK about removing hook_entity_preembed()? Now that we're not using placeholder rendering, there's no real reason to have two different hooks (hook_entity_embed_context_alter + hook_entity_preembed) to invoke prior to the embedded entity being rendered.

cs-shadow:
@davereid That definitely makes sense.

PR #161, opened 2015-08-13
Use embed module for preview and AJAX insertion.

davereid:


js/plugins/drupalentity/plugin.js
```diff
@@ -211,54 +211,4 @@
...
-   * Attaches or detaches behaviors, except the ones we do not want.
-   *
-   * @param {string} action
-   *   Either 'attach' or 'detach'.
-   * @param context
-   *   The context argument for Drupal.attachBehaviors()/detachBehaviors().
-   * @param settings
-   *   The settings argument for Drupal.attachBehaviors()/detachBehaviors().
-   */
-  function runEmbedBehaviors(action, context, settings) {
```
cs-shadow, line 224: If we're removing this function, where are we adding this functionality?

davereid, line 224: So I had already written the AJAX embedding abstraction (sorry @cs-shadow). I added it to https://github.com/drupal-media/entity_embed/pull/161. I might just merge that PR now and file follow-up issues for what's left to do...

cs-shadow, line 224: Sure, no problem.

PR #165, opened 2015-08-14
Separate out embed dialog methods.

davereid:


cs-shadow:
I like this, it makes the methods more easily readable and understandable.

PR #177, opened 2015-09-04
Added an entity embed service.

davereid:
I thought it would be good to separate out the actual rendering part to a service that could be overridden. The main benefit I was thinking of is having a constant for the default display plugin ID. But I could just as easily add that to `EntityEmbedDisplayManager`.


davereid:
I guess it comes down to, should we shove everything into one trait, or is it better to split out the parts responsible for rendering into its own thing.

PR #179, opened 2015-09-10
Update ckeditor plugin

davereid:
Trying to solve https://www.drupal.org/node/2554687

### DIFF
--- a/js/plugins/drupalentity/plugin.js
+++ b/js/plugins/drupalentity/plugin.js
@@ -25,8 +25,8 @@
     init: function (editor) {
       // Generic command for adding/editing entities of all types.
       editor.addCommand('editdrupalentity', {
-        allowedContent: 'drupal-entity[*]',
-        requiredContent: 'drupal-entity[*]',
+        allowedContent: 'drupal-entity[alt,title,data-*]',
+        requiredContent: 'drupal-entity[data-entity-type,data-embed-button]',
         modes: { wysiwyg : 1 },
         canUndo: true,
         exec: function (editor, data) {
@@ -81,9 +81,8 @@
 
       // Register the entity embed widget.
       editor.widgets.add('drupalentity', {
-        // Minimum HTML which is required by this widget to work.
-        allowedContent: 'drupal-entity[*]',
-        requiredContent: 'drupal-entity[*]',
+        allowedContent: 'drupal-entity[alt,title,data-*]',
+        requiredContent: 'drupal-entity[data-entity-type]',
 
         // Simply recognize the element as our own. The inner markup if fetched
         // and inserted the init() callback, since it requires the actual DOM
--- a/js/plugins/drupalentity/plugin.js
+++ b/js/plugins/drupalentity/plugin.js
@@ -154,7 +154,7 @@
         });
 
         editor.contextMenu.addListener(function(element) {
-          if (isEmbeddedEntityWidget(editor, element)) {
+          if (isEditableEntityWidget(editor, element)) {
             return { drupalentity: CKEDITOR.TRISTATE_OFF };
           }
         });
@@ -164,7 +164,7 @@
       editor.on('doubleclick', function (evt) {
         var element = getSelectedEmbeddedEntity(editor) || evt.data.element;
 
-        if (isEmbeddedEntityWidget(editor, element)) {
+        if (isEditableEntityWidget(editor, element)) {
           editor.execCommand('editdrupalentity');
         }
       });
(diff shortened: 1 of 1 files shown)


wimleers:
See https://www.drupal.org/node/2554687#comment-10315745 for the attributes aspect here.

But the custom element aspect, I don't know about that — we'll definitely need the guidance from the CKEditor developers for that. I'm sure that this will become a more and more common question, so it'd be in their best interest to answer it once here, or develop the best approach here, and then document it clearly in the Widgets documentation.
Pinging @Reinmar, @oleq, @wwalc and @fredck.

davereid:
Also, hooray, this breaks the 'View Source' and back. Need to investigate.

js/plugins/drupalentity/plugin.js
```diff
@@ -8,27 +8,25 @@
...
-        }
-      }
-
+      CKEDITOR.dtd['drupal-entity'] = {'#': 1};
+      CKEDITOR.dtd.$block['drupal-entity'] = 1;
+      CKEDITOR.dtd.$blockLimit['drupal-entity'] = 1;
+      // @todo It should be possible to display embedded entities inline...
+      //CKEDITOR.dtd.$inline['drupal-entity'] = 1;
+      CKEDITOR.dtd.$nonEditable['drupal-entity'] = 1;
+      CKEDITOR.dtd.$object['drupal-entity'] = 1;
```
davereid, line 22: I'm totally not sure if I'm doing any of the above correctly.

js/plugins/drupalentity/plugin.js
```diff
@@ -200,6 +197,18 @@
...
+   * Returns whether or not the given element is a editable drupalentity widget.
+   *
+   * This requires the widget to have the data-embed-button attribute present.
+   *
+   * @param {CKEDITOR.editor} editor
+   * @param {CKEDITOR.htmlParser.element} element
+   */
+  function isEditableEntityWidget (editor, element) {
+    return isEmbeddedEntityWidget(editor, element); //&& element.$.firstChild.attributes.hasOwnProperty('data-embed-button');
+  }
```
davereid, line 209: Feels like I should be able to just remove this if the editdrupalentity command requires the data-embed-button attribute in its requiredContent.

davereid:
Man, this is way out of my comfort zone.

js/plugins/drupalentity/plugin.js
Reinmar, line 22: Removing the above code (that loop) isn't correct. In the DTD you need to define a couple of things:
1. What your tag can contain (you did this with `CKEDITOR.dtd['drupal-entity'] = {'#': 1};` – it means that your tag can only contain a text).
2. Then, you need to define what other elements can be parents of your element. That was what the loop did. It said that every element that can contain a `div` will be able to contain your element (so, in other words – treat your element like a `div`).
3. Finally, there are some special objects which define additional semantics. The `$block` contains elements that should be treated as blocks (like `<p>`), `$blockLimit` is a block which cannot be freely split (like `<td>`) – to these two you should add your `drupal-entity`.

And that should be all, if your tag is later replaced by some widget. If it's not, then... well – you shouldn't be using non-HTML tags at all :P.

js/plugins/drupalentity/plugin.js
```diff
@@ -8,27 +8,25 @@
...
-      dtd['drupal-entity'] = {'#': 1};
-      // Register drupal-entity element as allowed child, in each tag that can
-      // contain a div element.
-      for (tagName in dtd) {
-        if (dtd[tagName].div) {
-          dtd[tagName]['drupal-entity'] = 1;
-        }
-      }
-
+      CKEDITOR.dtd['drupal-entity'] = {'#': 1};
```
mlewand, line 16: As far as comments in Drupa tickets go, you also might want to nest `drupal-entity` elements, so you also need to change this line to:
`CKEDITOR.dtd[ 'drupal-entity' ] = { '#': 1, 'drupal-entity': 1 };`

js/plugins/drupalentity/plugin.js
mlewand, line 22: Definitely you need to restore this for loop.

js/plugins/drupalentity/plugin.js
```diff
@@ -166,7 +163,7 @@
       editor.on('doubleclick', function (evt) {
         var element = getSelectedEmbeddedEntity(editor) || evt.data.element;
 
-        if (isEmbeddedEntityWidget(editor, element)) {
+        if (isEditableEntityWidget(editor, element)) {
           editor.execCommand('editdrupalentity');
```
mlewand, line 167: You could also call `evt.cancel()` here so, you're sure that nothing else will handle this event.

slashrsm:
https://www.drupal.org/node/2554687 was fixed. I think that this can be closed.

PR #186, opened 2015-10-15
#2593315: Remove hook_entity_embed_alter()

wimleers:
See https://www.drupal.org/node/2593315 for details.

It does not and can not work as advertised, and best practice is to use custom view modes.


davereid:
We are not going to remove the hook. Let's fix it instead.

PR #187, opened 2015-10-15
Improve/fix handling of bubbleable metadata in entity_embed

wimleers:
See https://www.drupal.org/node/2593379.

### DIFF
--- a/src/Plugin/Filter/EntityEmbedFilter.php
+++ b/src/Plugin/Filter/EntityEmbedFilter.php
@@ -12,6 +12,8 @@ use Drupal\Core\Entity\EntityManagerInterface;
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Render\BubbleableMetadata;
+use Drupal\Core\Render\RenderContext;
 use Drupal\entity_embed\EntityHelperTrait;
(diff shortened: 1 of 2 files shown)


src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -102,13 +104,24 @@ public function process($text, $langcode) {
             }
 
             $access = $entity->access('view', NULL, TRUE);
-            $access_metadata = CacheableMetadata::createFromObject($access);
-            $entity_metadata = CacheableMetadata::createFromObject($entity);
```
davereid, line 106: How is it ok to remove the metadata about the entity itself? Are we assuming that every field formatter that can be re-used to display an embedded entity correctly has this metadata in its render API array?

davereid, line 106: @wimleers ?

davereid, line 106: This also looks like we're losing the metadata from the access check.

wimleers, line 106: We can remove the second line because we are rendering the embedded entity, that will itself include the cacheability metadata of the entity.

wimleers, line 106: > This also looks like we're losing the metadata from the access check.
> Why?

davereid, line 106: LOL, because I couldn't see the very next line:
$result->addCacheableDependency($access);

PR #188, opened 2015-10-15
Update README

wimleers:
The necessary attributes need to be whitelisted also now, it doesn't make sense to first mention `<drupal-entity>` and then use `<div>`, and fix some consistency problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Allows any entity to be embedded using a WYSIWYG and text format.
 
 * Install and enable the module.
 * Enable the entity-embed filter 'Display embedded entities' for the desired text formats from the configuration page: '/admin/config/content/formats'.
-* Add ```<drupal-entity>``` to the 'Allowed HTML tags'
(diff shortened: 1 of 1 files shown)


Berdir:
Updating the README is not enough, you also need to update 
allowedContent: 'drupal-entity[_]',
requiredContent: 'drupal-entity[_]',

to
allowedContent: 'drupal-entity[data-_]',
requiredContent: 'drupal-entity[data-_]',

in plugin.js.

I was just about the create an issue for this on d.o, I guess we can do this here as well.

(data-\* is a lot easier than listing them all and should be safe?)

wimleers:
> (data-\* is a lot easier than listing them all and should be safe?)

It's easier for sure, but it's also extremely imprecise. I think it's much safer/healthier to be explicit.

wimleers:
This was fixed in https://www.drupal.org/node/2628150.

For updating the CKEditor plugin, we have https://www.drupal.org/node/2554687.

wimleers:
https://www.drupal.org/node/2628150 = #194.

PR #189, opened 2015-11-11
Decouple engine from UI

phenaproxima:
I have a use case wherein I would like to be able to slap drupal-entity tags into my CKEditor, but without using Entity Embed's UI at all, and without its button(s) appearing in the editor.

This PR decouples the "engine" part of Entity Embed -- i.e., the widget definition -- from the UI and command parts. They are split into two separate JS files -- engine.js and plugin.js -- and two separate plugins on the Drupal side (drupalentity and drupalentity_button).


mirodietiker:
That sounds really interesting and i think if entity_embed wants to offer nice helpers for all use cases we need this kind of separation.

slashrsm:
Great work! I agree with @mirodietiker. This change makes a lot of sense. I gave it a quick sanity check only as I don't have time to do a proper review ATM. I hope me or somebody else comes back to properly review and test asap.

webflo:
:+1:

balsama:
This is great - and likely a common use case. +1

davereid:
Please, I'd love to review this as well.

src/Plugin/CKEditorPlugin/DrupalEntity.php
```diff
@@ -8,7 +8,6 @@
 namespace Drupal\entity_embed\Plugin\CKEditorPlugin;
 
 use Drupal\editor\Entity\Editor;
-use Drupal\embed\EmbedButtonInterface;
 use Drupal\embed\EmbedCKEditorPluginBase;
```
davereid, line 11: Seems odd that this would still extend EmbedCKEditorPluginBase?

phenaproxima: The DrupalEntityButton plugin, which extends this one, is intended for CKEditor -- so it does need to extend EmbedCKEditorPluginBase. It's not the cleanest, but short of converting EmbedCKEditorPluginBase to a trait I'm not sure how else to approach that...

slashrsm, line 11: This is strange indeeed.

slashrsm:
I tested this PR and it seems to breaks CKEditor. 

Could we convert "engine" into a JS library? Would this allow us to use it in button plugin without extending "engine" class.

This PR is definitely a step in the right direction, but it needs some more love.

PR #193, opened 2015-11-30
Issue #2628226: EB Integration: Increase the default width of entity browser

webflo:
Depends on https://github.com/drupal-media/entity_browser/pull/120 and https://github.com/drupal-media/entity_browser/pull/121

Works best if the width is set to 100% inside of the entity browser settings.


css/entity_embed.dialog.css
```diff
@@ -0,0 +1,29 @@
+/**
+ * @file
+ * Styles for Entity Browser's modal windows.
+ */
+
+@media all and (max-width: 800px) {
+  .ui-dialog--narrow.entity-select-dialog {
+    min-width: 95%;
+    max-width: 95%;
```
neardark, line 9: just set width: 95% -- that's the same as min-width/max-width with same values.

webflo:
Closed in favour of https://github.com/drupal-media/entity_embed/pull/195

PR #195, opened 2015-12-03
Issue #2628226: EB Integration: Increase the default width of entity browser, Version 2

webflo:


slashrsm:
Rebased and merged. Thanks!

PR #200, opened 2015-12-11
Allow drupal-entity elements to be inline

phenaproxima:
I have a use case where it would be helpful for drupal-entity elements to be displayed inline. For instance, I may need to embed a tweet (from Media Entity Twitter) into a paragraph, floated to the left; or a video (through Media Entity Embeddable Video), floated to the right. Or perhaps an image (via Media Entity Image), inline with some text.

Right now the CKEditor DTD will not allow that, so this PR alters the DTD to allow drupal-entity elements as children of elements which can contain img elements.


slashrsm:
Creating d.o. issue so more people can find this and provide feedback: https://www.drupal.org/node/2640398

src/Plugin/CKEditorPlugin/DrupalEntity.php
```diff
@@ -25,6 +26,17 @@ class DrupalEntity extends EmbedCKEditorPluginBase {
   /**
    * {@inheritdoc}
    */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    $configuration['inline'] = $container->get('config.factory')
```
slashrsm, line 30: It would make sense to have this configurable per-button.

phenaproxima, line 30: I'm not sure it's going to be possible to make it configurable per button. The reason being that making drupal-entity elements inline requires changing the DTD of the editor instance, which will take effect for the entire editor, regardless of how many different embed buttons are loaded into it. It's intrinsically a per-editor thing, not a per-button thing. But maybe I'm misunderstanding you -- what did you have in mind?

js/plugins/drupalentity/plugin.js
```diff
@@ -18,9 +18,10 @@
       var dtd = CKEDITOR.dtd, tagName;
       dtd['drupal-entity'] = {'#': 1};
       // Register drupal-entity element as allowed child, in each tag that can
-      // contain a div element.
+      // contain a div or img element (if inline embedding is turned on in the
+      // settings).
       for (tagName in dtd) {
-        if (dtd[tagName].div) {
+        if (dtd[tagName].div || (dtd[tagName].img && editor.config.DrupalEntity_inline)) {
```
mirodietiker, line 24: I think we should consider on !!CKEDITOR.dtd.$inline[tagName] instead of img.

src/Plugin/CKEditorPlugin/DrupalEntity.php
mirodietiker, line 30: Oh wow that's quite some limitation of CKEditor then...

So if we need to limit this per button / entity type, we would need to have different tags. We can NOT depend on classes / attributes in the DTD.  But using multiple tags is also a mismatch to the current filter approach - it would need extra configuration and changing it contains risk stale data.

How about considering an autogenerated (optional?) "drupal-entity-TYPE" tag that would allow specific DTDs? Yeah quite some change / added complexity in the filter..

davereid:
I don't think this is a viable approach because CKEditor is horrible at supporting an element being block level and also inline at the same time. I believe we'll have to use two different tags, maybe `<drupal-entity>` and `<drupal-entity-inline>` ?

yannickoo:
@davereid so you could select the type of tag when inserting or editing?

phenaproxima:
@davereid, Can you be more specific as to the nature of CKEditor's sucktitude in this regard? I mean...couldn't we just have all drupal-entity tags be styled as inline-block, full stop?

bohemier:
Has there been any progress on this issue? I've been tinkering with this for past few days and agree that ckEditor does not play well with <drupal-entity> being inline (dtd won't allow them to be placed inside <p> tags, even inside divs they act weird,...)

PR #202, opened 2016-01-13
Make item viewable by anonymous

podarok:
There is no way to provide view access to anonymous user.
So here is a patch for that.


Berdir:
That seems wrong to me.

> There is no way to provide view access to anonymous user.

Why not? checking access for anon works perfectly fine, it's up to the entity to implement it correctly. I'm using it with nodes, files and polls, no problems.

slashrsm:
I agree with @Berdir. Which entity type are you trying to embed?

podarok:
@Berdir 
I was trying to embed Menu Link and got it not accessible for anon users

Berdir:
That's by design in core. See MenuLinkContentAccessControlHandler. It explicitly denies access and requires the admin permission for that. You can implement hook_ENTITY_TYPE_access() to change that, although I'm not sure what the implications of that and if anything becomes visible that you don't expect to be.

I think this can be closed.

Berdir:
Another approach might be to have a different type of embed display plugin that's specific for menu links and doesn't change access for anyone.

podarok:
actually this is something related to Entity Embed only.
Because changing core is not an option (via hooks -> security hole)
This stuff makes Entity Viewable by default(Via entity embed) and adds ability to simply override access via checkAccess() method within entity formatter without needs to define custom Entity Embed formatter.
I don't  think anybody needs such deep split by entity_embed and entity formatters level.
There are view modes for that.

davereid:
We cannot override this here, we must always check entity access. If Menu links needs a workaround, then we need a workaround, this is not the right solution.

PR #203, opened 2016-01-18
Fix removed redirects

podarok:
If module been used for embeding forms, redirect exception can't be executed.
This patch is a fix for that


src/Plugin/Filter/EntityEmbedFilter.php
```diff
@@ -117,6 +117,9 @@ public function process($text, $langcode) {
           }
         }
         catch(\Exception $e) {
+          if ($e instanceof \Drupal\Core\Form\EnforcedResponseException) {
+            throw $e;
```
slashrsm: Would this look nicer if we'd add another catch() block?

podarok:
@slashrsm done

slashrsm:
Looks good now. Thanks! 

Could you create an issue in https://www.drupal.org/project/issues/entity_embed, rebase and squash the two commits and reword commit message to reference issue # and contributors (yourself:)).

davereid:
Can you help explain this more? I don't really understand. An issue in the module issue queue would help.

podarok:
@davereid When you are inserting entity form with redirect via entity_embed - redirect created by throwing redirect Exception.
entity_embed doesn't respect that. The patch above fixes an issue

I'm talking about inserting Contact form(core module) via entity_embed.

podarok:
@slashrsm Yes, I can do that. Can you add me to co-maintainers on d.org?
Nickname there @podarok as well

slashrsm:
You don't need maintainer-ship to create an issue.

PR #204, opened 2016-01-26
#2634508 updated ckdeditor plugin to display entity after selection

kmoll:


slashrsm:
Merged with reworded commit message. Thanks!

PR #208, opened 2016-03-29
AJAX errors when embedding entities with Entity Browser

DevinCarlson:
With a fresh checkout of Entity Embed and Entity Browser, attempting to embed an entity using CKEditor fails with:

> Argument 1 passed to Drupal\entity_browser\Plugin\EntityBrowser\Display\Modal::displayEntityBrowser() must implement interface Drupal\Core\Form\FormStateInterface, none given, called in /entity_embed/src/Form/EntityEmbedDialog.php on line 211

It looks like Entity Browser's `displayEntityBrowser()` method must now be passed `$form_state`.

### DIFF
--- a/entity_embed.info.yml
+++ b/entity_embed.info.yml
@@ -7,3 +7,5 @@ dependencies:
   - editor
   - embed
   - filter
+test_dependencies:
+  - entity_browser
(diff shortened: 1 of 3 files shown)


slashrsm:
Merged. Thanks!

Drupal.org issue: https://www.drupal.org/node/2696261

PR #209, opened 2016-04-06
Adding ajax_base_page to stop ajax errors on embed

realitygaps:
Simple change which fixes the ajax errors when clicking an embed button


davereid:
This was filed as a patch to Drupal.org already, so it was fixed with https://github.com/drupal-media/entity_embed/pull/211

PR #210, opened 2016-04-07
Issue #2645458 by smaz, kmoll: update render trait to handle #item_at…

kmoll:
…tributes array element that image_formatter expects

### DIFF
--- a/src/EntityHelperTrait.php
+++ b/src/EntityHelperTrait.php
@@ -207,6 +207,11 @@ trait EntityHelperTrait {
       $build['#attributes']['data-caption'] = $context['data-caption'];
     }
 
+    // If this is an image, the image_formatter expects #item_attributes.
+    if ($build['#theme'] == 'image_formatter') {
+      $build['#item_attributes'] =  $build['#attributes'];
+    }
+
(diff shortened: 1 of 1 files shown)


davereid:
Would this be more appropriate to do in ImageFieldFormatter?

kmoll:
@davereid I would think that might be a better idea.  I am still a little confused why the ImageFieldFormatter uses #item_attributes over #attributes.  It seems to be the only place where this is the case.  I think if its possible, we update that.  If there are good reasons why its that way, then we would have to handle here, I think.  Let me know your thoughts.

davereid:
The core image field formatters require #item_attributes to be set, which is strange, but there's not much we can do about that from Entity Embed's side of things, we can only ensure the correct values are set.

slashrsm:
Unfortunately we can't move this as we add align, caption, ... classes after formatter plugin has returned the render array. Looking at test fails now.

slashrsm:
Attempt to fix test: https://www.drupal.org/node/2645458#comment-11117123

davereid:
Closing as this was fixed via Drupal.org

PR #212, opened 2016-04-13
Ensure the caption value is double-encoded for proper XSS filtering as an attribute on the embed.

davereid:
https://www.drupal.org/node/2688427


davereid:
I could use some help testing this:
1. With the change/patch, embed a new entity with an caption which contains HTML tags (`This caption is <strong>really</strong> great!`)
2. Test to make sure the caption appears correctly as HTML in the WYSIWYG preview when using multiple text formats.
3. Right click the embedded entity in the WYSIWYG, Select 'Edit entity', and confirm that the caption is the same value with the raw HTML tags.

davereid:
@balsama Not sure if you could help make this testing happen? ^

davereid:
Nevermind, I was finally able to fix the JS error I was getting and confirmed this is working as expected now!

PR #216, opened 2016-04-14
Issue #2554687 by Dave Reid: Add allowedContent to the CKEditor button registration, and updated requiredContent settings.

davereid:


davereid:
https://www.drupal.org/node/2554687 is fixed.

PR #218, opened 2016-04-14
Issue #2706051 by Dave Reid: Require UUIDs for embedding with the WYSWIYG and deprecate the data-entity-id attribute.

davereid:
https://www.drupal.org/node/2706051

### DIFF
--- a/src/Form/EntityEmbedDialog.php
+++ b/src/Form/EntityEmbedDialog.php
@@ -132,14 +132,13 @@ class EntityEmbedDialog extends FormBase {
     $entity_element += array(
       'data-entity-type' => $embed_button->getTypeSetting('entity_type'),
       'data-entity-uuid' => '',
-      'data-entity-id' => '',
       'data-entity-embed-display' => 'entity_reference:entity_reference_entity_view',
       'data-entity-embed-settings' => array(),
       'data-align' => '',
       'data-caption' => '',
     );
     $form_state->set('entity_element', $entity_element);
-    $form_state->set('entity', $this->loadEntity($entity_element['data-entity-type'], $entity_element['data-entity-uuid'] ?: $entity_element['data-entity-id']));
+    $form_state->set('entity', $this->loadEntity($entity_element['data-entity-type'], $entity_element['data-entity-uuid']));
 
     if (!$form_state->get('step')) {
       // If an entity has been selected, then always skip to the embed options.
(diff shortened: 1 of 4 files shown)


src/Plugin/EmbedType/Entity.php
```diff
@@ -232,6 +232,10 @@ protected function getEntityTypeOptions() {
         if (!$this->entityTypeManager->getDefinition($entity_type_id)->hasViewBuilderClass()) {
           unset($options[$group][$entity_type_id]);
         }
+        // Filter out entity types that do not support UUIDs.
```
wimleers, line 235: I was gonna say:

> This is wrong because there always is one, see `\Drupal\Core\Entity\EntityInterface::uuid`!

But that's apparently not true:

```
   * @return string|null
   *   The UUID of the entity, or NULL if the entity does not have one.
   */
  public function uuid();
```

:(

So +1 for this.

src/Form/EntityEmbedDialog.php
```diff
@@ -369,10 +368,6 @@ public function buildEmbedStep(array $form, FormStateInterface $form_state) {
       '#type' => 'hidden',
       '#value' => $entity_element['data-entity-type'],
     );
-    $form['attributes']['data-entity-id'] = array(
-      '#type' => 'hidden',
-      '#value' => $entity_element['data-entity-id'],
-    );
     $form['attributes']['data-entity-uuid'] = array(
       '#type' => 'hidden',
       '#value' => $entity_element['data-entity-uuid'],
```
wimleers, line 373: This is a hidden form element. It gets its value not from a validate or submit callback. It gets it from `$entity_element`. Which I guess works because `$entity_element` is recalculated every time an auto-complete is submitted?

If so, please document this.

However, I still wonder if this _actually_ works when you start to link to an entity? (Or perhaps my mention of autocompletes is completely off even, I've been looking at LinkIt lately, so my memory of `entity_embed` is probably warped.)

davereid, line 373: Yeah, all the attributes get re-calculated on AJAX due to checking from state input.

wimleers, line 373: So, perhaps a quick comment documenting this? To prevent future puzzlement? :)

wimleers, line 373: And ideally a test that verifies the response that is sent after clicking the submit button.

slashrsm:
There is an updated patch that builds on top of this pull in https://www.drupal.org/node/2706051. Let's finish it there.

PR #219, opened 2016-04-14
Issue #2706059 by Dave Reid: Filter out empty attributes.

davereid:
https://www.drupal.org/node/2706059

### DIFF
--- a/src/Form/EntityEmbedDialog.php
+++ b/src/Form/EntityEmbedDialog.php
@@ -8,6 +8,7 @@
 namespace Drupal\entity_embed\Form;
 
 use Drupal\Component\Utility\Html;
+use Drupal\Component\Utility\Unicode;
 use Drupal\Core\Ajax\AjaxResponse;
 use Drupal\Core\Ajax\CloseModalDialogCommand;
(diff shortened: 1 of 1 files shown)


src/Form/EntityEmbedDialog.php
```diff
@@ -737,6 +736,11 @@ public function submitEmbedStep(array &$form, FormStateInterface $form_state) {
         $values['attributes']['data-entity-embed-settings'] = Json::encode($values['attributes']['data-entity-embed-settings']);
       }
 
+      // Filter out empty attributes.
+      $values['attributes'] = array_filter($values['attributes'], function($value) {
```
slashrsm, line 740: array_filter($values['attributes']) won't do?

davereid, line 740: It would filter out an alt text or caption value of "0", so I figured strlen was safer in this case.

slashrsm, line 740: Oh joys of PHP! I guess "0" isn't the most relevant value, but I understand.

slashrsm:
Merged rebased branch. Rebased commit is: 6cff3f3

PR #220, opened 2016-04-24
Issue #2712111 by slashrsm, paranojik: Outdated cached version of embedded entity

paranojik:
Problem was that the data attributes were added to the embedded entity itself. This caused the same data attributes to appear everywhere the same view mode was used/displayed on the site. That also meant that the cached version was not invalidated until the embedded entity's cache was cleared.

To prevent all mentioned problems we wrap the embedded entity and append the data attributes to the wrapper element.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Entity Embed can be installed via the [standard Drupal installation process](htt
   * Enable the 'Display embedded entities' filter.
   * Drag and drop the 'E' button into the Active toolbar.
   * If the text format uses the 'Limit allowed HTML tags and correct faulty HTML' filter, ensure the necessary tags and attributes are whitelisted: add ```<drupal-entity data-entity-type data-entity-uuid data-entity-id data-view-mode data-entity-embed-display data-entity-embed-settings data-align data-caption data-embed-button>``` to the 'Allowed HTML tags' setting. (Will happen automatically after https://www.drupal.org/node/2554687.)
+  * If you're using both the 'Align images' and 'Caption images' filters make sure the 'Align images' filter is run before the 'Caption images' filter in the **Filter processing order** section. (Explanation: Due to the implementation details of the two filters it is important to execute them in the right sequence in order to obtain a sensible final markup. In practice this means that the alignment filter has to be run before the caption filter, otherwise the alignment class will appear inside the <figure> tag (instead of appearing on it) the caption filter produces.)
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


davereid:
It would help me to understand what's actually wrong if we had a test case that failed first to prove this is fixing something.

davereid:
Would this also be related to https://github.com/drupal-media/entity_embed/pull/187 which is the fix for the bubbleable metadata?

slashrsm:
> Would this also be related to #187 which is the fix for the bubbleable metadata?

I don't think this is related. 

> It would help me to understand what's actually wrong if we had a test case that failed first to prove this is fixing something.

Check issue summary in https://www.drupal.org/node/2712111. We found another problem while working on the patch. Will describe it in the issue summary now.

wimleers:
> It would help me to understand what's actually wrong if we had a test case that failed first to prove this is fixing something.

+100

wimleers:
The filter that renders embedded entities must associate the necessary cacheability metadata. It sounds like that's not happening.

This is therefore basically a duplicate of https://www.drupal.org/node/2593379, which has PR #187. This only covers one of the many potential side-effects/consequences. The lack of test coverage for this is what's the real problem: `EntityEmbedFilterTest` only tests if something works at all, not if it is updated when necessary. We should at least test:
- that if an embedded entity is modified, the host entity's rendered output is also updated
- that if a tag on an embedded entity is added or removed, the host entity's rendered output is also updated
- that an embedded entity that becomes unpublished, it is also no longer rendered by the host entity's rendered output, because it's no longer accessible … but only for anonymous users

wimleers:
(Cross-posted from https://www.drupal.org/node/2712111#comment-11155929. These split conversations are mind-bending BTW.)

slashrsm:
> This is therefore basically a duplicate of https://www.drupal.org/node/2593379, which has PR #187. 

I am not sure about this. https://www.drupal.org/node/2593379 is about metadata from the embedded entity not being bubbled up. This is about entity_embed specific attributes being added to the part of the render array that belongs to the embedded entity and them being cached as part of it. 

In order to reproduce this embed a node using teaser view mode and use align/caption. Then navigate to the front page and inspect markup of the same entity when appearing on it's own (not embedded). You will see embed-specific attributes being added to it (data-align, data-caption), which clearly do not belong there. 

Will work on test that will prove that, but it might take me few days.

davereid:
https://www.drupal.org/node/2712111 has been fixed.

